PR #8, opened 2018-07-04
Add the dependencies to the setup.py

valo:
Also renamed the main package to sanpy


tspenov:
Just a question: is it required to rename it to `sanpy` ?

valo:
Yes. Otherwise `import sanpy` should be `import san`

PR #9, opened 2018-07-05
Remove gql lib dependancy

tspenov:


san/graphql.py
```diff
@@ -1,20 +1,18 @@
...
+def execute_gql(gql_query_str):
+    headers = {}
     if ApiConfig.api_key:
-        kwargs['headers'] = {'authorization':  "Apikey {}".format(ApiConfig.api_key)}
-
-    return Client(
-            transport=RequestsHTTPTransport(**kwargs),
-            fetch_schema_from_transport=True
-        )
+        headers = {'authorization':  "Apikey {}".format(ApiConfig.api_key)}
```
IvanIvanoff: Should we add the authorization header if there is no apikey provided?

PR #10, opened 2018-07-05
Add default params for queries

tspenov:
Default values for parameters:
* `from_date`: `datetime.now() - 365 days`
* `to_date`: `datetime.now()`
* `interval`: `'1d'`

It is possible to call just like that:
```python
daa = san.get("daily_active_addresses/santiment")
prices = san.get("prices/santiment")
```


san/sanbase_graphql.py
```diff
@@ -1,4 +1,9 @@
 import iso8601
+import datetime
+
+DEFAULT_TO_DATE = datetime.datetime.now().isoformat()
+DEFAULT_FROM_DATE = (datetime.datetime.now() - datetime.timedelta(days=365)).isoformat()
```
valo: I think this value will be calculated when the library loads and won't change from there, so if the scripts runs for more than 1 day the default will point to yesterday. May be we can have these defaults as lambdas and calculate them when they are used?

PR #14, opened 2018-08-09
Add endpoints

LucoLe:
Adds endpoints for exchange_funds_flow, erc20_exchange_funds_flow,
social_volume_projects, social_volume.

Adds documentation for the new endpoints.


README.md
```diff
@@ -275,3 +260,192 @@ datetime
 2018-06-04 00:00:00+00:00  0.07947329054545459  602.5116327272724
 
 ```
+
+### Exchange funds flow
+
+Fetch the difference between the tokens that were put in minus the tokens that were taken out from an exchange for a given slug in the selected time period.
```
valo: Let's use the words `deposited` and `withdrawn` as these clearly indicate what we are talking about.

README.md
```diff
@@ -275,3 +260,192 @@ datetime
...
+* exchangeDiffBtc - The difference between the bought and the sold tokens in BTC: exchangeInBtc - exchangeOutBtc
+* percentDiffExchangeDiffBtc - The percent difference between exchangeDiffBtc for the current period minus the exchangeDiffBtc for the previous period based on exchangeDiffBtc for the current period: (exchangeDiffBtc for current period - exchangeDiffBtc for previous period) * 100 / abs(exchangeDiffBtc for current period)
+* exchangeVolumeBtc - The volume of all tokens in and out for the given period in BTC: exchangeInBtc + exchangeOutBtc
+* percentDiffExchangeVolumeBtc - The percent difference between exchangeVolumeBtc for the current period minus the exchangeVolumeBtc for the previous period based on exchangeVolumeBtc for the current period: (exchangeVolumeBtc for current period - exchangeVolumeBtc for previous period) * 100 / abs(exchangeVolumeBtc for current period)
+
+```python
+
+# The "/" at the end of the first argument is important. If you omit it you'll get an error.
+prices = san.get(
+    "erc20_exchange_funds_flow/",
```
valo: The trailing slash here is a bit weird. Can we go without it?

README.md
```diff
@@ -275,3 +260,192 @@ datetime
...
+                              fundsFlow
+datetime
+2018-04-16 10:02:19+00:00      0.000000
+2018-04-16 10:17:18+00:00      0.000000
+2018-04-16 12:03:51+00:00      0.000000
+2018-04-16 13:50:24+00:00   -208.797310
+
+```
+
+### ERC20 Exchange Funds Flow
```
valo: I am wondering if we should name this a bit differently. It is a kind of an overview of the exchange funds flow + as we add more blockchains it won't cover only ERC20, but all blockchains. May be `Exchange Funds Flow Overview`?

README.md
```diff
@@ -275,3 +260,192 @@ datetime
...
+```
+
+### Social Volume Projects
+
+Fetch a list of slugs for which there is social volume data.
+
+```python
+
+# The "/" at the end of the first argument is important. If you omit it you'll get an error.
+prices = san.get("social_volume_projects/")
```
valo: The trailing slash looks weird to me. What do you think?

san/sanbase_graphql.py
```diff
@@ -1,9 +1,16 @@
 import iso8601
 import datetime
 
-DEFAULT_TO_DATE = lambda: datetime.datetime.now()
-DEFAULT_FROM_DATE = lambda: datetime.datetime.now() - datetime.timedelta(days=365)
+
+def DEFAULT_TO_DATE(): return datetime.datetime.now()
```
tspenov: Why do you remove the lambda ? It is there so the date is calculated when it is used, not when the library is loaded.

README.md
tspenov: Yes, we should make it work without the trailing slash. I think it will be better

tspenov:
Note: this one should be merged after these metrics are on PROD

README.md
```diff
@@ -275,3 +260,192 @@ datetime
...
+
+### Exchange funds flow
+
+Fetch the difference between the tokens that were put in minus the tokens that were taken out from an exchange for a given slug in the selected time period.
+
+```python
+
+prices = san.get(
+    "exchange_funds_flow/santiment",
+    from_date="2018-04-16T10:02:19Z",
```
tspenov: I think we can make the examples only with dates: `from_date="2018-04-16"` - it will look easier to use ?

README.md
```diff
@@ -275,3 +260,192 @@ datetime
...
+
+```
+
+### Social Volume
+
+Fetch a list of mentions count for a given project and time interval.
+
+Arguments description:
+
+* endpoint - social_volume/dragonchain
```
tspenov: endpoint - social_volume/project_slug ?

README.md
tspenov: I think it's not `prices` here anymore ?

README.md
```diff
@@ -275,3 +260,192 @@ datetime
...
+* social_volume_type - one of the following:
+    1. PROFESSIONAL_TRADERS_CHAT_OVERVIEW
+    2. TELEGRAM_CHATS_OVERVIEW
+    3. TELEGRAM_DISCUSSION_OVERVIEW
+
+    It is used to select the source of the mentions count.
+
+```python
+
+prices = san.get(
```
tspenov: here it's not `prices` - maybe `mentions` or just `social` or no var :)

san/sanbase_graphql.py
tspenov: I see but why is the change then ?

LucoLe: Haha that was a linter change, to be honest IMO let's change the casing and keep it as a function but if you prefer I can change it back to a lambda.

README.md
```diff
@@ -23,15 +23,15 @@ If the account used for generating the api key has enough SAN tokens, the api ke
...
+* `query`: Available queries can be found in section: [Available metrics](#available-metrics)
 * `slug`: A list of projects with their slugs, names, etc. can be fetched like this:
 
 ```python
 import san
 san.get("projects/all")
 ```
 
-```
+```bash
```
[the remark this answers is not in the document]
LucoLe: There are is no significant difference but I'll change it to `console`

README.md
```diff
@@ -275,3 +260,192 @@ datetime
...
+### ERC20 Exchange Funds Flow
+
+Fetch the exchange funds flow for all ERC20 projects in the given interval.
+
+Arguments description:
+
+* from_date - a string representation of datetime value according to the iso8601 standard, e.g. "2018-04-16T10:02:19Z"
+* to_date - a string representation of datetime value according to the iso8601 standard, e.g. "2018-05-23T10:02:19Z"
+
+Fields description:
```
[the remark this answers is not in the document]
LucoLe: they are used in camelCase and as you can see in the output they are returned in camelCase as well so I guess we should leave it this way.

README.md
LucoLe: Is it really that big of a difference if it's `2018-04-16T00:00:00Z` or `2018-04-16T10:02:19Z` and yes I'd like to leave it with time for the moment. This way we have places where we show only date and datetime.

README.md
LucoLe: I agree but let's do this in a separate task because this will need changes in tech indicators, grafana, sanbase and sanpy.

README.md
```diff
@@ -275,3 +260,190 @@ datetime
...
+
+Example result:
+
+```console
+
+                              fundsFlow
+datetime
+2018-04-16 10:02:19+00:00      0.000000
+2018-04-16 10:17:18+00:00      0.000000
+2018-04-16 12:03:51+00:00      0.000000
```
valo: This example looks weird. Should these be daily values? Why they are zeros?

README.md
valo: 😢 seems like we need to do quite a bit of planning around naming things that will be included in the public APIs...

PR #17, opened 2018-08-14
Fix the README so it's properly displayed in pypi

LucoLe:
Removes the `console` from the fenced code blocks in order to be
properly displayed on pypi.

Adds a note that the user needs to set the api key and have at least
1000 SAN tokens in their account in order to use the social_volume.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ## Installation
 
-```console
+```
 pip install sanpy
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -407,7 +407,7 @@ Example result:
 
 ### Social Volume
 
-Fetch a list of mentions count for a given project and time interval.
+Fetch a list of mentions count for a given project and time interval. For this metric you'll need to set the api key and have at least 1000 SAN tokens in your account.
```
tspenov: can we add a link to the apikey configuration section ?

PR #19, opened 2018-08-22
Add topic_search to sanpy

LucoLe:
Adds the topic_search endpoint to the sanpy library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -452,14 +452,14 @@ datetime
 
 ### Topic search
 
-Returns lists with the mentions of the search phrase grouped by source. The results are in two formats - the messages themselves and the data for building graph representation of the result. For this metric you'll need to set the [api key](#configuration) and have some SAN tokens in your account.
+Returns lists with the mentions of the search phrase from the selected source. The results are in two formats - the messages themselves and the data for building graph representation of the result. For this metric you'll need to set the [api key](#configuration) and have some SAN tokens in your account.
 
 Arguments description:
 
 * A string in the format "topic_search/`fields`" where `fields` is one of the following:
     1. "messages"
-    2. "charts_data"
-* sources - an array of enum values that specify which sources should be searched. The sources are:
+    2. "chart_data"
+* source - one of the following:
     1. TELEGRAM
     2. PROFESSIONAL_TRADERS_CHAT
     3. REDDIT
@@ -472,7 +472,7 @@ Arguments description:
 
 topic_search = san.get(
     "topic_search/messages",
-    sources="TELEGRAM, PROFESSIONAL_TRADERS_CHAT, REDDIT",
+    source="TELEGRAM",
     search_text="btc moon",
     from_date="2018-08-01T12:00:00Z",
     to_date="2018-08-15T12:00:00Z",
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -448,4 +448,48 @@ datetime
...
+
+                                                                chartsData  \
+professionalTradersChat  [{'mentionsCount': 170, 'datetime': '2018-08-0...
+reddit                   [{'mentionsCount': 316, 'datetime': '2018-08-0...
+telegram                 [{'mentionsCount': 219, 'datetime': '2018-08-0...
+
+                                                                  messages
+professionalTradersChat  [{'text': 'damn if eth does break tendline btc...
+reddit                   [{'text': 'In the last 2 bull markets (April-M...
+telegram                 [{'text': 'Btc dominance increasing', 'datetim...
```
valo: I am wondering if we should split these into 2 separate requests, like `topic_search/mentions` and `topic_search/messages`. I am not sure that you need both all the time. Also I think it will be much easier to work with the data if the index is datetime and the columns are `professionalTradersChat`, `reddit` and `telegram` (this is for the mentions).

About the format of the messages: may be again we make an index on datetime with columns `text` and `type` which will be one of `professionalTradersChat`, `reddit` to `telegram`

What do you think?

LucoLe: It's not a problem to split this method to two requests for messages and chart data. And I really don't have an opinion about this.

Having an index of the datetime and merging multiple queries is too complicated IMO and I'd prefer to keep it the way we have it now. Anyway, this is going to be a bigger change and if we decide to go that way I won't be able to do it today.

README.md
```diff
@@ -110,7 +110,7 @@ Below are described some available metrics and are given examples for fetching a
 
 ### Daily Active Addresses
 
-This metric includes the number of unique addresses that participated in the transfers of given token during the day.
+This metric includes the number of unique addresses that participated in the transfers of given token during the day. For this metric you'll need to set the [api key](#configuration) and have some SAN tokens in your account.
```
tspenov: I don't think you need API key for this metric. Also not for `Token aging (burn rate)`, `TV`.... Can you check this ?

LucoLe: I made a change in `sanbase` so these endpoints require the user to be logged in. This is agreed with @valo

valo:
Please, check with the backend team. While you were on vacation we opened the DAA and the dev activity APIs.

README.md
tspenov: But here for example: https://api.santiment.net/graphiql?variables=%7B%7D&query=query%20%7B%0A%20%20transactionVolume(%0A%20%20%20%20slug:%20%22santiment%22,%0A%20%20%20%20from:%20%222018-01-01%2016:00:00Z%22,%0A%20%20%20%20to:%20%222018-06-05%2016:00:00Z%22,%0A%20%20%20%20interval:%20%2215m%22)%20%7B%0A%20%20%20%20%20%20transactionVolume,%0A%20%20%20%20%20%20datetime%0A%20%20%20%20%7D%0A%7D%0A
You can take TV without being logged in.

tspenov: But you can still get TV for: https://api.santiment.net/graphiql?variables=%7B%7D&query=query%20%7B%0A%20%20transactionVolume(%0A%20%20%20%20slug%3A%20%22etheroll%22%2C%0A%20%20%20%20from%3A%20%222018-07-01%2016%3A00%3A00Z%22%2C%0A%20%20%20%20to%3A%20%222018-07-05%2016%3A00%3A00Z%22%2C%0A%20%20%20%20interval%3A%20%2215m%22)%20%7B%0A%20%20%20%20%20%20transactionVolume%2C%0A%20%20%20%20%20%20datetime%0A%20%20%20%20%7D%0A%7D%0A
Its just in the interval [Now() - 3months,  Now() - 1day]

tspenov: Santiment slug - no timeframe restriction.
Also for DAA and Github Activity we allow historical data now, but not realtime. (that was from last week)
For TV, Burn Rate, .. - you can still get this timeframe [Now() - 3months, Now() - 1day] without apikey.

PR #26, opened 2018-10-08
Fix the exchangeFundsFlow query - rename the field queried

IvanIvanoff:


IvanIvanoff:
@tspenov done. My markdown formatter also changed some other things

tspenov:
Oh, and also the version has to be changed if we want to deploy it to pypi

tspenov:
👍

PR #28, opened 2018-10-11
Add integration testing

tspenov:
Check whether all queries to sanbase return result

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ with open("README.md", "r") as fh:
 
 setuptools.setup(
     name="sanpy",
-    version="0.0.20",
+    version="0.0.21",
     author="Santiment",
     author_email="admin@santiment.net",
     description="Package for Santiment API access with python",
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ with open("README.md", "r") as fh:
 
 setuptools.setup(
     name="sanpy",
-    version="0.1.20",
+    version="0.1.0",
     author="Santiment",
     author_email="admin@santiment.net",
     description="Package for Santiment API access with python",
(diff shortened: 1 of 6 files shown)


IvanIvanoff:
Should these tests always run in the CI as they require internet and sanbase?

setup.py
```diff
@@ -6,7 +6,7 @@
 
 setuptools.setup(
     name="sanpy",
-    version="0.0.20",
+    version="0.0.21",
```
Partysun: What is the logic behind the version system, do we use in sanpy?
Usually, the third value is the bugs/fixes.

tspenov: @Partysun For now we're just bumping the version so we can upload to pypi. But this is a good suggestion - we may start utilizing what you suggested: `major.minor.bugfixes`. So this PR should bump the minor. Also adding a CHANGELOG file

IvanIvanoff: https://semver.org

setup.py
```diff
@@ -6,7 +6,7 @@
 
 setuptools.setup(
     name="sanpy",
-    version="0.0.20",
+    version="0.1.20",
```
IvanIvanoff: If you bump the second number you need to set the last to 0 -> `0.1.0`

tspenov:
@IvanIvanoff There is some tradeoff with running tests that require external dependancy but I think having this kind of integration test will be more beneficial. We can try it on and see how it goes.

setup.py
tspenov: ups yes, indeed

valo:
Can we run these tests with a special command and not run them by default when building the PRs? This way we will be able to define a separate build job which runs these once a day?

My concern is that if sanbase goes down all the builds will stop working and the master branch will be blocked...

tspenov:
The command for running integration tests is:
```
python setup.py nosetests -a integration
```

Issue #34, opened 2018-10-25
Returning empty dataframe on response 200 with errors


tspenov:
The actual errors should be returned

Issue #33, opened 2018-10-25
ohlcv queries cannot be batched


tspenov:
Add note in the README

PR #35, opened 2018-10-25
Handle status 200 with 'errors' in the json response, update readme

tspenov:
Fixes https://github.com/santiment/sanpy/issues/34 and https://github.com/santiment/sanpy/issues/33

* if one query and it return errors - show the errors
* if we have batched queries - show errors only when all queries return errors, otherwise return empty dataframe when the query returns error

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,6 +297,8 @@ datetime
 
 ### open/close prices with volume and marketcap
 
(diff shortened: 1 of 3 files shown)


tspenov:
Although for batched queries this is not right bacause 5 queries can return data and the 6th can have errors.

PR #37, opened 2018-10-26
telegram chat/discussion explanation added

ivanklimuk:
Improve the API docs about the difference between telegram chats and telegram discussion


ivanklimuk:
@tspenov done

Issue #39, opened 2018-11-07
fix default to_date to be the end of the day


tspenov:
`to_date=2018-10-29` defaults to `2018-10-29T00:00:00`. It will be better to default to `2018-10-29T23:59:59` so it returns all the data for the day.

```
san.get('transaction_volume/omisego', from_date='2018-10-28', to_date='2018-10-29')
                           transactionVolume
datetime
2018-10-28 00:00:00+00:00       3.226938e+05
2018-10-29 00:00:00+00:00       1.281221e+03
```

```
>>> san.get('transaction_volume/omisego', from_date='2018-10-28', to_date='2018-10-29T00:00:00')
                           transactionVolume
datetime
2018-10-28 00:00:00+00:00      322693.810945
2018-10-29 00:00:00+00:00        1281.220900
```

```
san.get('transaction_volume/omisego', from_date='2018-10-28', to_date='2018-10-29T23:59:59')
                           transactionVolume
datetime
2018-10-28 00:00:00+00:00       3.226938e+05
2018-10-29 00:00:00+00:00       9.269391e+05
```

PR #40, opened 2018-11-11
Change default to date

tspenov:
should fix : https://github.com/santiment/sanpy/issues/39

### DIFF
--- a/san/sanbase_graphql.py
+++ b/san/sanbase_graphql.py
@@ -317,4 +317,4 @@ def _format_to_date(datetime_obj_or_str):
         datetime_obj_or_str = datetime_obj_or_str.isoformat()
 
     dt = iso8601.parse_date(datetime_obj_or_str) + datetime.timedelta(hours=23, minutes=59, seconds=59)
-    dt.isoformat()
+    return dt.isoformat()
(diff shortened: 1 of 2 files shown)


san/sanbase_graphql.py
```diff
@@ -287,8 +287,15 @@ def _format_return_fields(return_fields):
...
     if isinstance(datetime_obj_or_str, datetime.datetime):
         datetime_obj_or_str = datetime_obj_or_str.isoformat()
 
     return iso8601.parse_date(datetime_obj_or_str).isoformat()
+
+def _format_to_date(datetime_obj_or_str):
+    if isinstance(datetime_obj_or_str, datetime.datetime):
+        datetime_obj_or_str = datetime_obj_or_str.isoformat()
+
+    dt = iso8601.parse_date(datetime_obj_or_str) + datetime.timedelta(hours=23, minutes=59, seconds=59)
```
valo: Why is the end of the day taken here?

tspenov: Seems to me that the end of the day is more appropriate here otherwise it includes this day but without the full data for the day. You can check the examples here: https://github.com/santiment/sanpy/issues/39

PR #45, opened 2019-01-23
Add warning outdated version

tspenov:
warning like this:
```
>>> import san
/Users/tspenov/devel/sanpy/san/__init__.py:20: UserWarning: The package sanpy is out of date. Your version is 0.2.1, the latest is 0.3.1.
  warn(warning)
>>>
```

### DIFF
--- a/san/__init__.py
+++ b/san/__init__.py
@@ -1,3 +1,20 @@
 from .get import get
 from .batch import Batch
 from .api_config import ApiConfig
+import pkg_resources
+import requests
+import json
+from warnings import warn
+
+def get_latest(project):
+    url = 'https://pypi.python.org/pypi/%s/json' % (project)
+    response = requests.get(url).text
+    return json.loads(response)['info']['version']
+
+project = 'sanpy'
+current_version = pkg_resources.require(project)[0].version
+latest = get_latest(project)
+
+if current_version != latest:
+    warning = 'The package %s is out of date. Your version is %s, the latest is %s.' % (project, current_version, latest)
+    warn(warning)
--- a/san/__init__.py
+++ b/san/__init__.py
@@ -5,7 +5,7 @@ import pkg_resources
 import requests
 import json
 from warnings import warn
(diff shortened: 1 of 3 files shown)


IvanIvanoff:
@tspenov can we add some instruction how to upgrade it? I am not familiar with the python ecosystem and if `pip` is the only and standard way to upgrade packages.

san/__init__.py
```diff
@@ -1,3 +1,20 @@
...
 from .api_config import ApiConfig
+import pkg_resources
+import requests
+import json
+from warnings import warn
+
+def get_latest(project):
+    url = 'https://pypi.python.org/pypi/%s/json' % (project)
+    response = requests.get(url).text
+    return json.loads(response)['info']['version']
```
valo: May be we can make this check async and if the network call fails, we just ignore it? That should fix the load performance and allow to load the module when there is no network connection (it is a bit useless without it, but still it could be annoying to fail on import I guess)

IvanIvanoff:
Also I think it is running the network calls in the CI, right?

san/__init__.py
tspenov: Yeah, added the check when there is no network

tspenov:
@IvanIvanoff I think `pip` is pretty much the standard way - there is `easy_install` and `Anaconda` I don't think they are used so much. Also added in the README

tspenov:
@IvanIvanoff for sure it is used at least with running the tests

san/__init__.py
tspenov: @valo also added the check to be async

setup.py
```diff
@@ -21,6 +21,7 @@
         'pandas',
         'numpy',
         'requests',
-        'iso8601'
+        'iso8601',
+        'setuptools'
```
tspenov: unfortunately for the check I need `pkg_resources` which is part of `setuptools` - so I should include it as a install dependancy

PR #49, opened 2019-05-02
Add 'Token Age Consumed' and 'Average Token Age Consumed in Days'

balresch:
- let 'burn_rate' point to TokenAgeConsumed in GraphQL, so that 'burnRate' can be removed from GraphQL.
- add 'token_age_consumed', to replace 'burn_rate' in the long run
- add 'average_token_age_consumed_in_days'
- update readme accordingly

### DIFF
--- a/san/sanbase_graphql.py
+++ b/san/sanbase_graphql.py
@@ -15,9 +15,17 @@ QUERY_MAPPING = {
         'query': 'dailyActiveAddresses',
         'return_fields': ['datetime', 'activeAddresses']
     },
-    'burn_rate': {
-        'query': 'burnRate',
-        'return_fields': ['datetime', 'burnRate']
+    'burn_rate': { # to be removed
+        'query': 'tokenAgeConsumed',
+        'return_fields': ['datetime', 'tokenAgeConsumed']
+    },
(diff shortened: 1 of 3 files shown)


san/sanbase_graphql.py
```diff
@@ -15,9 +15,17 @@
         'query': 'dailyActiveAddresses',
         'return_fields': ['datetime', 'activeAddresses']
     },
-    'burn_rate': {
-        'query': 'burnRate',
-        'return_fields': ['datetime', 'burnRate']
+    'burn_rate': { # to be removed
+        'query': 'tokenAgeConsumed',
+        'return_fields': ['datetime', 'tokenAgeConsumed']
```
tspenov: Here the return field is changed which will also break existing code. I would prefer the old query to stay and just add a deprecation warning when burn_rate is invoked : https://github.com/santiment/sanpy/pull/49/files#diff-0f5457262476aa06c3885256259a709dR69
example:
https://github.com/santiment/sanpy/blob/1f48d7dd01bdec929feabe6dc62e5c24798fb95c/san/__init__.py#L25

tspenov:
🎉 🎉 🎉

PR #50, opened 2019-05-05
Adding missing metrics

balresch:
token velocity, token circulation, realized value, mvrv, nvt, daily active deposits added, readme and version updated

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ with open("README.md", "r") as fh:
 
 setuptools.setup(
     name="sanpy",
-    version="0.4.3",
+    version="0.5.0",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -6,7 +6,7 @@
 
 setuptools.setup(
     name="sanpy",
-    version="0.4.2",
```
tspenov: we can bump to 0.5.0 with so much new metrics included

tspenov:
💙 💙 💙 Great work!!!

setup.py
tspenov: updated

PR #56, opened 2019-07-06
Add gas used

spiderjako:


README.md
```diff
@@ -854,6 +855,35 @@ datetime
...
 ```
 
+### Gas Used
+
+Returns used Gas by a blockchain. When you send tokens, interact with a contract or do anything else on the blockchain, you must pay for that computation. That payment is calculated in Gas.
+
+
+```python
+result = san.get(
```
kalkov: Remove the `result =`, it is not used anywhere

README.md
```diff
@@ -854,6 +855,35 @@ datetime
 
 ```
 
+### Gas Used
+
+Returns used Gas by a blockchain. When you send tokens, interact with a contract or do anything else on the blockchain, you must pay for that computation. That payment is calculated in Gas.
+
```
kalkov: Mention that only ethereum is supported.

PR #57, opened 2019-07-06
Add miners balance

spiderjako:


README.md
```diff
@@ -854,6 +855,35 @@ datetime
...
 ```
 
+### Miners Balance
+
+Returns miner balances over time. Currently only ETH is supported.
+
+
+```python
+result = san.get(
```
kalkov: Remove `result =` assignment, it is not used

README.md
```diff
@@ -854,6 +855,35 @@ datetime
...
+    interval="1d"
+)
+```
+
+Example result:
+
+
+```
+                                balance
+datetime
```
IvanIvanoff: `datetime` and `balance` can go on the same line

kalkov:
This branch was created from gas-used instead of master, so it contains changes from both branches.

PR #60, opened 2019-07-08
Add top_holders_percent_of_total_supply metric

spiderjako:


README.md
```diff
@@ -854,6 +855,34 @@ datetime
...
+    to_date="2019-04-11"
+)
+```
+
+Example Result:
+
+
+```
+                           inExchanges  inTopHoldersTotal  outsideExchanges
+datetime
```
IvanIvanoff: these should go on one line

README.md
```diff
@@ -854,6 +855,33 @@ datetime
...
+
+### Top Holders Percent Of Total Supply
+
+Returns the top holders' percent of total supply - in exchanges, outside exchanges and combined.
+
+[Premium metric](#premium-metrics)
+
+```python
+san.get(
+    "top_holders_percent_of_total_supply/ethereum",
```
IvanIvanoff: @tspenov @kalkov how should we proceed here where we have time restrictions? For `slug` based metics we can show `sentiment`s free data but here we do not have this option

kalkov: We can add to the premium metrics section 
```markdown
```
something like:

PR #65, opened 2019-07-10
Add test for ordinary functions

spiderjako:


san/tests/test_sanbase_integration.py
```diff
@@ -73,6 +73,25 @@ def test_erc20_projects():
...
+
+    for query in queries:
+        if query in ["gas_used", "miners_balance", "mining_pools_distribution"]:
+            slug_to_test = "ethereum"
+        elif query in ["historical_balance", "top_holders_percent_of_total_supply", "social_dominance"]:
+            continue
+        else:
+            slug_to_test = params["project_slug"]
+        result = san.get(
+            "{}/{}".format(query, slug_to_test),
```
IvanIvanoff: `slug_to_test` is undefined if we're in the `elif` case. What happens in this case?

IvanIvanoff: My bad, `continue` will skip over this

san/tests/test_sanbase_integration.py
```diff
@@ -73,6 +73,25 @@ def test_erc20_projects():
...
+def test_ordinary_function():
+    queries = sanbase_graphql.QUERY_MAPPING.keys()
+
+    for query in queries:
+        if query in ["gas_used", "miners_balance", "mining_pools_distribution"]:
+            slug_to_test = "ethereum"
+        elif query in ["historical_balance", "top_holders_percent_of_total_supply", "social_dominance"]:
+            continue
+        else:
+            slug_to_test = params["project_slug"]
```
IvanIvanoff: You can leave this test but IMO explicit > implicit. So it will be better to have multiple functions where you explicitly write down exactly what query is tested. This way you can also break into multiple tests `test_social_data.py`, `test_blockchain_data.py`, `test_prices.py`, etc.

san/tests/test_sanbase_integration.py
```diff
@@ -73,6 +73,25 @@ def test_erc20_projects():
     assert len(result) > 0
     assert len(result[result.slug == "bitcoin"]) == 0
 
+@attr('integration')
+def test_ordinary_function():
```
kalkov: A single test case for testing more than one query is not a good idea, also having a loop + conditional in a test. 

You can declare some constants
```python
METRICS_USING_ETHEREUM = ["gas_used", "miners_balance", "mining_pools_distribution"]
METRICS_USING_SANTIMENT = ["historical_balance", "top_holders_percent_of_total_supply", "social_dominance"]
```
and use [nose test generators](https://nose.readthedocs.io/en/latest/writing_tests.html#test-generators)

That way we will have test case for every function and simple iteration w/o conditional.

san/tests/test_sanbase_integration.py
```diff
@@ -73,6 +73,25 @@ def test_erc20_projects():
...
+            continue
+        else:
+            slug_to_test = params["project_slug"]
+        result = san.get(
+            "{}/{}".format(query, slug_to_test),
+            from_date=params["from_date"],
+            to_date=params["to_date"],
+            interval=params["interval"]
+        )
+        assert len(result.index) >=1
```
IvanIvanoff: This asserts can test a little more things. You can test that you have the expected number of returned data (if from-to is 5 days long time range and interval is "1d" you expect 5 results). Also you can test that the results have some properties (many of the metrics are non negative numbers)

san/sanbase_graphql.py
```diff
@@ -94,18 +94,6 @@
     'mining_pools_distribution':{
         'query': 'miningPoolsDistribution',
         'return_fields': ['datetime', 'other', 'top10', 'top3']
-    },
```
kalkov: Why these are removed?

san/tests/test_sanbase_integration.py
```diff
@@ -27,7 +38,7 @@ def test_batched_queries_equal_format():
         )
 
     result = batch.execute()
-
+    print(result)
```
kalkov: leftover

san/tests/test_sanbase_integration.py
```diff
@@ -10,12 +10,23 @@
...
+        from_date=params['from_date'],
+        to_date=params['to_date'],
+        interval=params['interval']
+        )
+    return result
+
 @attr('integration')
 def test_batched_queries_equal_format():
-    queries = sanbase_graphql.QUERY_MAPPING.keys()
+    queries = sanbase_graphql.QUERY_MAPPING.keys() - METRICS_USING_ETHEREUM
```
kalkov: Why are metrics using ethereum not included?

san/tests/test_sanbase_integration.py
spiderjako: You are right, I did as you asked and started making the tests for every function, not one test for all functions

san/tests/test_sanbase_integration.py
spiderjako: You're right, the tests should cover more criterias, not only the number of returned results. I did the new tests with a few new assertions, but will do more also.

san/tests/test_sanbase_integration.py
spiderjako: You're right about the single test case - all queries. I've made new tests, which explicitly test out one query at a time. (did not use nose test generators though)

san/sanbase_graphql.py
spiderjako: Because when I've started making those functions for the queries, I thought, that every new function should be there, but it turned out, that it wasn't needed. Moreover, the test for the batch for equal format queries fail, because those 3 functions aren't equal format

san/tests/test_sanbase_integration.py
spiderjako: Because I test out every function with equal format, which can use "santiment" as a slug. Then I did another for cycle, so that I cover the ones that use "ethereum" also.

san/tests/test_sanbase_integration.py
```diff
@@ -73,6 +91,77 @@ def test_erc20_projects():
...
+    for query in METRICS_USING_ETHEREUM:
+        result = ordinary_function_maker(query, "ethereum")
+
+        assert len(result.index) >= 1
+
+@attr('integration')
+def test_token_age_consumed():
+    result = ordinary_function_maker('token_age_consumed')
+
+    for row in result['tokenAgeConsumed']:
```
kalkov: The assertions can go into `ordinary_function_maker` it can be renamed to `test_ordinary_function` and the call will be as simple as:

```python
def test_token_age_consumed():
    test_ordinary_function('token_age_consumed', 'tokenAgeConsumed')
```

spiderjako: You're right, I've moved and renamed the function accordingly

PR #66, opened 2019-07-10
Add price_volume_difference metric

spiderjako:


README.md
```diff
@@ -1026,6 +1027,32 @@ datetime                   inExchanges  inTopHoldersTotal  outsideExchanges
 2019-04-11 00:00:00+00:00     7.975260          13.296356          5.321096
 ```
 
+## Price Volume Difference
+
+Fetch the price-volume difference technical indicator for a given ticker, display currency and time period. This indicator measures the difference in trend between price and volume, specifically when price goes up as volume goes down.
```
IvanIvanoff: ticker -> slug

README.md
```diff
@@ -1026,6 +1027,32 @@ datetime                   inExchanges  inTopHoldersTotal  outsideExchanges
...
+
+```python
+san.get(
+    "price_volume_difference/santiment",
+    from_date="2019-04-18",
+    to_date="2019-04-23",
+    interval="1d",
+    currency="USD",
+    size=0
```
IvanIvanoff: Wouldn't it be better to use `size=0` and not being able to change it? Or at least have a default value for it + description

san/sanbase_graphql.py
```diff
@@ -280,6 +280,31 @@ def top_holders_percent_of_total_supply(idx, slug, **kwargs):
...
+    kwargs = _transform_query_args(**kwargs)
+
+    query_str = """
+    query_{idx}: priceVolumeDiff (
+        slug: \"{slug}\",
+        from: \"{from_date}\",
+        to: \"{to_date}\",
+        interval: \"{interval}\",
+        currency: \"{currency}\",
+        size: 0
```
IvanIvanoff: Do not pass this param - the graphql server will handle it ok

PR #67, opened 2019-07-10
Revert "Update readme"

spiderjako:
Reverts santiment/sanpy#64


IvanIvanoff:
Let's fix the version in a PR that get's merged after this

PR #68, opened 2019-07-11
Add eth_top_transactions metric

spiderjako:


README.md
```diff
@@ -1026,6 +1027,36 @@ datetime                   inExchanges  inTopHoldersTotal  outsideExchanges
...
+    transaction_type="ALL"
+)
+```
+
+Example result:
+
+```
+
+datetime                                                         fromAddress                                          toAddress                                            trxHash  trxValue                                                                                                                                                     
+2019-04-29 21:33:31+00:00  {'address': '0xe76fe52a251c8f3a5dcd657e47a6c8d...  {'address': '0x45d6275d9496bc61b5b072a75089be4...  0x776cd57382456adb2f9c22bc2e06b4ddc9937b073e7f...    100.00
```
IvanIvanoff: Same comments as the ones in the token top transactions

san/get.py
```diff
@@ -16,4 +30,14 @@ def get(dataset, **kwargs):
     gql_query = "{\n" + gql_query + "\n}"
     res = execute_gql(gql_query)
 
+    if query in QUERY_PATH_MAP:
+        res = shorten_path(res, QUERY_PATH_MAP[query])
+        if query == 'eth_top_transactions':
```
IvanIvanoff: This is a special case and is awkward to have it in the generic `get` function. Currently I think we do not have a better place to place it be we need to implement it.

san/get.py
```diff
@@ -1,12 +1,26 @@
...
+    the first is the first key in the result of the query, the last - the key that we need the data from.
+
+    The returned value is a shortened dictionary, from which the query can easily return the needed data.
+    """
 CUSTOM_QUERIES = {
     'ohlcv': 'get_ohlcv',
 }
 
+QUERY_PATH_MAP = {
+    'eth_top_transactions': ['query_0', 'ethTopTransactions']
```
IvanIvanoff: why `query_0`? We should have the automatic assignment of the numbers, right? If so, this query won't be always be with the `0` index. You can test by using `batch` and batching multiple `ethTopTransactions` -  you can add this as a test

san/get.py
IvanIvanoff: Possible solution is to be able to attach transform functions to the queries that will run before the result is returned.   Having by default the idempotent function (`id(x) = x`) and only changing where it's needed is a more extendable way.

IvanIvanoff: That being said, if we have such transform functions that work in a generic way they can be executed here. Pseudo code:
```python
if query in QUERY_PATH_MAP:
   res = shorten_path(res, QUERY_PATH_MAP[query])
   transform_fn = get_transform_fn(query)
   apply transform_fn to res
```

san/batch.py
```diff
@@ -23,7 +23,8 @@ def execute(self):
         idxs = sorted([int(k.split('_')[1]) for k in res.keys()])
         for idx in idxs:
             key = "query_{}".format(idx)
```
IvanIvanoff: this var is unused

spiderjako: Yes, sorry for that, I forgot, that it isn't needed anymore, when I have the transforming function. I've just removed it.

san/query.py
```diff
@@ -14,6 +18,9 @@ def parse_dataset(dataset):
     left, _separator, right = dataset.partition("/")
     return [left, right]
 
+@add_transform
+def transform_query(idx, query, result):
+    return convert_to_datetime_idx_df(result['query_'+str(idx)])
```
IvanIvanoff: This is called from within the decorator

PR #69, opened 2019-07-11
Add token_top_transactions metric

spiderjako:


README.md
```diff
@@ -1026,6 +1027,31 @@ datetime                   inExchanges  inTopHoldersTotal  outsideExchanges
...
+    limit=5,
+)
+```
+
+Example result:
+
+```
+
+datetime                                                         fromAddress                                          toAddress                                            trxHash      trxValue
+2019-04-21 13:51:59+00:00  {'address': '0x1f3df0b8390bb8e9e322972c5e75583...  {'address': '0x5eaae5e94995202c2183956369b1700...  0xdbced935b09dd0b9747d004f6242eb1e31b4770ae994...  166674.00000
```
IvanIvanoff: Something is missing from from the address here (`...`).
Also, let's map `toAddress` to just address and not the json `'adderss': <addr>, `isExchange`: <boolean>`.
What we could do is to add 2 additional columns `toAddressIsExchange` and `fromAddressIsExchange` with boolean values

san/get.py
```diff
@@ -1,12 +1,21 @@
 import san.sanbase_graphql
+import operator
 from san.graphql import execute_gql
 from san.pandas_utils import convert_to_datetime_idx_df
 from san.query import get_gql_query, parse_dataset
+from functools import reduce
+
+def get_by_path(root, items):
+    return reduce(operator.getitem, items, root)
```
IvanIvanoff: Add documentation and example what this function is doing

README.md
```diff
@@ -1026,6 +1027,43 @@ datetime                   inExchanges  inTopHoldersTotal  outsideExchanges
 2019-04-11 00:00:00+00:00     7.975260          13.296356          5.321096
 ```
 
+### Token Top Transactions
+
+Top transactions for the token of a given project
+
+For the successful execution of the function, the usage of an external function - get_by_path - is needed. The arguments for it are the result from the execution of the query and a list from two items, the first of which is 'query_0' and the second - the key, to which we need to cut the path, and the returned result is a shortened list of the returned values from the query.
```
IvanIvanoff: You should put this in the function's documentation. All implementation details should not be known by the documentation that the users will read

README.md
```diff
@@ -1026,6 +1027,43 @@ datetime                   inExchanges  inTopHoldersTotal  outsideExchanges
...
+Example result:
+
+```
+datetime                                                  fromAddress  fromAddressInExchange  ...                                            trxHash      trxValue                                              ...                                                                 
+2019-04-21 13:51:59+00:00  0x1f3df0b8390bb8e9e322972c5e75583e87608ec2                  False  ...  0xdbced935b09dd0b9747d004f6242eb1e31b4770ae994...  166674.00000
+2019-04-28 07:43:38+00:00  0x0a920bfdf7f9777a32699aeb7a44d857c1ad7d1e                  False  ...  0x5f2214d34bcdc3298799c1a58ee73e4d95ee94310e78...   33181.82279
+2019-04-28 07:53:32+00:00  0x868074aab18ea32e0103e9b3c10cc0a958eeb324                  False  ...  0x90bd286da38a2bf6de0aa2415e74bfd1400982168e9d...   33181.82279
+2019-04-26 14:38:45+00:00  0x876eabf441b2ee5b5b0554fd502a8e0600950cfa                   True  ...  0xe45b86f415e9300afb1caa2f2abcb5e965f22acc617a...   28999.64023
+2019-04-30 15:17:28+00:00  0x876eabf441b2ee5b5b0554fd502a8e0600950cfa                   True  ...  0xc85892b9ef8c647957583cb9fb1f44a7dd89ee91410d...   20544.42975
+```
```
IvanIvanoff: the `toAddress` data is missing. Also shorten the addresses and the hash like this:
```
datetime                          fromAddress    fromAddressIsExchange            trxHash      trxValue
2019-04-21 13:51:59+00:00  0x1f3df0b8390bb8...                   False   0xdbced935b09dd0  166674.00000
2019-04-28 07:43:38+00:00  0x0a920bfdf7f977...                   False   0x5f2214d34bcdc3   33181.82279
2019-04-28 07:53:32+00:00  0x868074aab18ea3...                   False   0x90bd286da38a2b   33181.82279
2019-04-26 14:38:45+00:00  0x876eabf441b2ee...                    True   0xe45b86f415e930   28999.64023
2019-04-30 15:17:28+00:00  0x876eabf441b2ee...                    True   0xc85892b9ef8c64   20544.42975
```

README.md
IvanIvanoff: Like this: https://www.datacamp.com/community/tutorials/docstrings-python#second-head

san/get.py
```diff
@@ -16,4 +30,14 @@ def get(dataset, **kwargs):
     gql_query = "{\n" + gql_query + "\n}"
     res = execute_gql(gql_query)
 
+    if query in QUERY_PATH_MAP:
+        res = shorten_path(res, QUERY_PATH_MAP[query])
+        if query == 'token_top_transactions':
+            for column in res:
+                column['fromAddressInExchange'] = column['fromAddress']['isExchange']
```
IvanIvanoff: `is` instead of `in`

README.md
spiderjako: You're right, it looks better this way as you said.

san/get.py
spiderjako: Did as you asked, but the function is renamed to be "path_to_data" and is moved to the "transform.py" file

README.md
spiderjako: Did so for the "path_to_data" function, which was "get_by_path" before

README.md
spiderjako: You're right, I've fixed the README. It should look normal now.

san/get.py
spiderjako: Did as you said.

IvanIvanoff:
let's first merge #68 as this one is kinda based on it

san/sanbase_graphql.py
```diff
@@ -333,11 +333,11 @@ def eth_top_transactions(idx, slug, **kwargs):
...
-            ethTopTransactions (
-                from: \"{from_date}\",
-                to: \"{to_date}\",
-                limit: {limit},
-                transactionType: {transaction_type}
+        ethTopTransactions (
+            from: \"{from_date}\",
+            to: \"{to_date}\",
+            limit: {limit},
+            transaction_type: {transaction_type}
```
IvanIvanoff: use camelCase here. It is true that our backend understands both but it's standard (or at least more comment) for the client to use camelCase

spiderjako: My mistake, when I was merging the conflicts it got messed up and I deleted it and then thought that its name was with underscore. Fixed it

tspenov:
I think we should bump the version to `0.6.0` cause this is the last of the list of metrics that will add for now ? @ivan @kalkov @spiderjako

PR #70, opened 2019-07-12
Add news metric

spiderjako:


README.md
```diff
@@ -1026,6 +1027,45 @@ datetime                   inExchanges  inTopHoldersTotal  outsideExchanges
...
 
+### News
+
+Returns the news for given word.
+
+Arguments description:
+
+  *  tag - Project name, ticker or other crypto related words.
+  *  from - a string representation of datetime value according to the iso8601 standard, e.g. "2018-04-16T10:02:19Z"
+  *  to - a string representation of datetime value according to the iso8601 standard, e.g. "2018-04-16T10:02:19Z"
```
IvanIvanoff: The example is using just date (`2019-04-18`) which is not valid ISO8601.

README.md
```diff
@@ -1026,6 +1027,45 @@ datetime                   inExchanges  inTopHoldersTotal  outsideExchanges
...
+  *  size - size limit of the returned results
+
+[Premium metric](#premium-metrics)
+
+```python
+san.get(
+    "news/bitcoin",
+    from_date="2019-04-18",
+    to_date="2019-07-11",
+    size=10
```
IvanIvanoff: Let's have smaller `size` in the example and instead show the whole URL

san/sanbase_graphql.py
```diff
@@ -280,6 +280,30 @@ def top_holders_percent_of_total_supply(idx, slug, **kwargs):
...
+        tag: \"{tag}\",
+        from: \"{from_date}\",
+        to: \"{to_date}\",
+        size: {size}
+    ){{
+        datetime,
+        description,
+        sourceName,
+        title,
+        url
```
IvanIvanoff: Let's rearrange the params in the order:
1. datetime
2. title
3. description
4. source
5. url

README.md
```diff
@@ -1026,6 +1027,42 @@ datetime                   inExchanges  inTopHoldersTotal  outsideExchanges
 2019-04-11 00:00:00+00:00     7.975260          13.296356          5.321096
 ```
 
+### News
+
+Returns the news for given word.
+
+Arguments description:
+
+  *  tag - Project name, ticker or other crypto related words.
```
IvanIvanoff: nitpick: only `tag`'s description starts with a capital letter

spiderjako: Made it into a lowercase letter

san/sanbase_graphql.py
spiderjako: Done

IvanIvanoff:
let's merge #68 first so the diff here is smaller

README.md
```diff
@@ -43,6 +43,8 @@ Santiment API python client.
     - [Top Holders Percent of Total Supply](#top-holders-percent-of-total-supply)
     - [History Twitter Data](#history-twitter-data)
     - [Price Volume Difference](#price-volume-difference)
+    - [Ethereum Top Transactions](#ethereum-top-transactions)
```
tspenov: why do we have this one in `news` PR ?

spiderjako: I added the metric to the already updated master (the Ethereum Top Transactions was the last PR)

san/pandas_utils.py
```diff
@@ -2,7 +2,7 @@
 
 
 def convert_to_datetime_idx_df(data):
-    df = pd.DataFrame(data)
+    df = pd.DataFrame(data, columns=data[0].keys())
```
tspenov: Does that change anything for this result and also for other queries results or it is the default one?

san/transform.py
```diff
@@ -35,4 +36,11 @@ def eth_top_transactions_transform(data):
         column['toAddressIsExchange'] = column['toAddress']['isExchange']
         column['fromAddress'] = column['fromAddress']['address']
         column['toAddress'] = column['toAddress']['address']
+
+
+def news(data):
```
IvanIvanoff: `news_transform` otherwise it won't be called

IvanIvanoff: Also add documentation to this module on how it works - that you need to define `<query>_transform` function

san/pandas_utils.py
spiderjako: This is needed for the news metric, because otherwise, when the dataframe is being made from the "transformed" query, it turns it back to alphabetical order, which is not what we want for the news metric. The ordinary metrics are okay with this, but I you are right in the sense, that the transformed queries are ordered differently. I've just updated the eth_top_transactions transform function, so that it also reorders the result there.

spiderjako: If this column reordering isn't done, then the DataFrame will be automatically set to be in alphabetical order

san/transform.py
spiderjako: @IvanIvanoff Just saw the second comment. The documentation should be about the news_transform or the transform functions in general ?

IvanIvanoff: About the whole module

san/pandas_utils.py
tspenov: for news you want them to be the same as they are in the graphql query request ?
```
        datetime,
        title,
        description,
        sourceName,
        url
```

tspenov: I think this is feature of python 3.6+: https://docs.python.org/3.6/whatsnew/3.6.html#whatsnew36-compactdict

spiderjako: By default it is description->sourceName->title, I wanted it reordered

tspenov: I think it can't be depended upon (since lower versions user won't get the same order) , but also doesn't hurt

IvanIvanoff: won't it result in a runtime error if you execute it with python < 3.6?

san/transform.py
```diff
@@ -32,7 +37,18 @@ def transform_query_result(idx, query, data):
...
     for column in data:
         column['fromAddressIsExchange'] = column['fromAddress']['isExchange']
-        column['toAddressIsExchange'] = column['toAddress']['isExchange']
         column['fromAddress'] = column['fromAddress']['address']
+        column['toAddressIsExchange'] = column['toAddress']['isExchange']
         column['toAddress'] = column['toAddress']['address']
+        column['toAddress'] = column.pop('toAddress')
+        column['toAddressIsExchange'] = column.pop('toAddressIsExchange')
+        column['trxHash'] = column.pop('trxHash')
+        column['trxValue'] = column.pop('trxValue')
```
tspenov: there is no return here ?

spiderjako: You're right, I will add it now, though it is kind of a void function

README.md
kalkov: This one should not be here, maybe you need to rebase against master?

san/transform.py
kalkov: These also

san/pandas_utils.py
tspenov: it will work, just the order of keys won't be preserved in lower versions

README.md
spiderjako: Ethereum Top Transactions is the last PR request, doesn't that mean that it is already in master (the last PR is merged) ?

san/transform.py
```diff
@@ -1,9 +1,14 @@
...
+ functions, which reorder or make different dictionaries in general.
+"""
 import operator
 from san.pandas_utils import convert_to_datetime_idx_df
 from functools import reduce
 
 QUERY_PATH_MAP = {
-    'eth_top_transactions': ['ethTopTransactions']
+    'eth_top_transactions': ['ethTopTransactions'],
+    'news': []
```
tspenov: why using empty array here - should `news` be in the `QUERY_PATH_MAP` at all ?
I think we should change this because that should be 2 separate conditions:
```python
    if query + '_transform' in globals():
        result = path_to_data(idx, query, data)
        globals()[query + '_transform'](result)
    else:
        result = data['query_' + str(idx)]
``` 
1. if if query + '_transform' in globals()
2. if query in QUERY_PATH_MAP
and we won't need `news` key here

spiderjako: I thought that the idea was to use as little conditions as possible, that's why I made the 'news' key in QUERY_PATH_MAP, so I wouldn't have to use another condition.

tspenov: Also the `news_transform` won't be necessary if this reworked :?

spiderjako: Why not? I need it in order to reorder the different columns to not be in alphabetical order

spiderjako: The other case is that I remove the upgrade to pandas_utils file and just leave the news be as it is, in that case I really won't need an extra function

tspenov: I think condition in the general function will save defining useless paths in the QUERY_PATH_MAP
 
```python
def transform_query_result(idx, query, data):
    """
    If there is a transforming function for this query, then the result is
    passed for it for another transformation
    """
    if query in QUERY_PATH_MAP:
        result = path_to_data(idx, query, data)
    else:
        result = data['query_' + str(idx)]

    if query + '_transform' in globals():
        globals()[query + '_transform'](result)

    return convert_to_datetime_idx_df(result)
```

and we can remove `news_transform` and `news` key from `QUERY_PAT_MAP`.

san/transform.py
```diff
@@ -21,18 +25,33 @@ def transform_query_result(idx, query, data):
...
+        column['toAddressIsExchange'] = column['toAddress']['isExchange']
         column['toAddress'] = column['toAddress']['address']
+        column['toAddress'] = column.pop('toAddress')
+        column['toAddressIsExchange'] = column.pop('toAddressIsExchange')
+        column['trxHash'] = column.pop('trxHash')
+        column['trxValue'] = column.pop('trxValue')
+    return data
+
+
+def news_transform(data):
```
tspenov: can we make the transform in a more functional programming way without `pop`.
something like this:
```python
def news_transform(data):
    return list(map(lambda column: {
       'title': column['title'],
       'description': column['description'],
       ...
    }, data))
```

tspenov: I think that will be easier to grasp :?

spiderjako: You are right, I've made both of the functions this way now

san/tests/test_get.py
```diff
@@ -33,8 +29,8 @@ def test_get(mock):
         to_date="2018-06-05",
         interval="1d"
     )
-    expected_df = convert_to_datetime_idx_df(api_call_result['query_0'])
```
IvanIvanoff: you could have left the naming

san/transform.py
```diff
@@ -3,14 +3,14 @@
  functions, which reorder or make different dictionaries in general.
 """
 import operator
+import copy
```
IvanIvanoff: this is unused here

spiderjako: You're right, it's removed now

san/tests/test_get.py
spiderjako: You're right, so I just did, thanks

PR #73, opened 2019-07-15
Add eth-top-transactions metric. Add transformation for query result.

spiderjako:


san/transform.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    With this function we jump straight onto the key from the dataframe, that we want and start from there
+    """
+
+def add_transform(function):
+    def decorator(idx, query, data):
+        if query in QUERY_PATH_MAP:
+            print(underscore_to_camelcase(query))
+            result = path_to_data(idx, query, data)
+            globals()[query+'_transform'](result)
```
IvanIvanoff: why `globals`?

san/transform.py
```diff
@@ -0,0 +1,32 @@
...
+def path_to_data(idx, query, data):
+    return reduce(operator.getitem, ['query_'+str(idx),]+QUERY_PATH_MAP[query], data)
+    """
+    With this function we jump straight onto the key from the dataframe, that we want and start from there
+    """
+
+def add_transform(function):
+    def decorator(idx, query, data):
+        if query in QUERY_PATH_MAP:
+            print(underscore_to_camelcase(query))
```
IvanIvanoff: remove

spiderjako: Yes, I just saw, that I should remove a few things, among them are a few imports in some files

san/transform.py
```diff
@@ -0,0 +1,30 @@
...
+
+def path_to_data(idx, query, data):
+    return reduce(operator.getitem, ['query_'+str(idx),]+QUERY_PATH_MAP[query], data)
+    """
+    With this function we jump straight onto the key from the dataframe, that we want and start from there
+    """
+
+def add_transform(function):
+    def decorator(idx, query, data):
+        if query+'_transform' in globals():
```
tspenov: Maybe one comment here is needed that if transforming function for the query exists -> the result is passed to it for another transformation.

san/query.py
```diff
@@ -16,5 +19,10 @@ def parse_dataset(dataset):
     return [left, right]
 
 
+@add_transform
+def transform_query(idx, query, result):
```
IvanIvanoff: Now that I think this will better fit the `transform.py` module and probably with a different name - `transform_query_result` or just `transform` as it's not the query that's transformed, but the result :?

san/transform.py
```diff
@@ -0,0 +1,30 @@
...
+}
+
+def path_to_data(idx, query, data):
+    return reduce(operator.getitem, ['query_'+str(idx),]+QUERY_PATH_MAP[query], data)
+    """
+    With this function we jump straight onto the key from the dataframe, that we want and start from there
+    """
+
+def add_transform(function):
+    def decorator(idx, query, data):
```
IvanIvanoff: My original idea when suggesting decorators was that every transform function will be decorated as to show what query it transforms. Though the currently implemented way is with dynamically choosing the transform function based on its string name.

In this case the decorator could be removed as it's not doing any work - we can move the `transform_query` function here, call it just `transform_result` (check the other comment) and leave the definition of `decorator` as its implementation (+ call the `convert` on the result instead of `function`)

tspenov: Makes sense. Decorator is not needed. The solution with checking globals whether a transform function exists is sufficient. For the name: I think `transform_query_result` - cause the transformation is based on the query name

san/batch.py
```diff
@@ -23,8 +22,8 @@ def execute(self):
 
         idxs = sorted([int(k.split('_')[1]) for k in res.keys()])
         for idx in idxs:
-            key = "query_{}".format(idx)
-            df = convert_to_datetime_idx_df(res[key])
+            query = self.queries[idx][0].split("/")[0]
+            df = transform_query_result(idx, query, res)
```
IvanIvanoff: can you somehow hardcode the expected result? if the transformation is not applied when it has to this test will still pass as transformations won't be applied on both the result and the expected result

san/transform.py
spiderjako: I've just put the comment right before the "if" statement

san/query.py
spiderjako: I renamed the function and moved it into transform.py

san/transform.py
spiderjako: Removed the decorator and renamed the function, it really fits it more than just transform_query

san/batch.py
spiderjako: Just finished up hardcoding the tests expected results

san/transform.py
```diff
@@ -0,0 +1,30 @@
...
+
+QUERY_PATH_MAP = {
+    'eth_top_transactions': ['ethTopTransactions']
+}
+
+def path_to_data(idx, query, data):
+    return reduce(operator.getitem, ['query_'+str(idx),]+QUERY_PATH_MAP[query], data)
+    """
+    With this function we jump straight onto the key from the dataframe, that we want and start from there
+    """
```
IvanIvanoff: Is this comment for `path_to_data`? If so - it should be placed before the code, i.e. above `return`

san/transform.py
```diff
@@ -0,0 +1,30 @@
...
+
+def transform_query_result(idx, query, data):
+    # If there is a transforming function for this query, then the result is passed for it for another transformation
```
IvanIvanoff: keep the lines relatively short - this can go on 2 lines

san/transform.py
tspenov: I am wondering why autopep8 doesn't catch those ? @spiderjako do you use it `autopep8  --in-place --aggressive *.py` before commiting ?

IvanIvanoff: autopep8 indeed formats some changes from this PR, though it does not change the comment. Formatting tools usually ignore comments

spiderjako: I've used the autopep8, yes, but I guess, it didn't catch the comment, as Ivan suggested.

spiderjako: And yes, the comment is for path_to_data

san/transform.py
```diff
@@ -0,0 +1,31 @@
...
+
+def path_to_data(idx, query, data):
+    """
+    With this function we jump straight onto the key from the dataframe, that we want and start from there
+    """
+    return reduce(operator.getitem, ['query_'+str(idx),]+QUERY_PATH_MAP[query], data)
+
+def transform_query_result(idx, query, data):
+    # If there is a transforming function for this query, then the result is
+    # passed for it for another transformation
```
IvanIvanoff: Keep the heredoc comment style (""" .... """)

san/tests/test_get.py
```diff
@@ -79,6 +93,385 @@ def test_error_response(mock):
...
+    # expected
+    mock.return_value = TestResponse(status_code=200, data=deepcopy(expected))
+
+    res = san.get(
+        "daily_active_addresses/santiment",
+        from_date="2018-06-01",
+        to_date="2018-06-05",
+        interval="1d"
+    )
+    expected = convert_to_datetime_idx_df(expected['query_0'])
```
IvanIvanoff: `expected_df` - reusing var names is confusing, even more when it also changes the type (dict -> pandas data frame)

san/tests/test_get.py
```diff
@@ -79,6 +93,385 @@ def test_error_response(mock):
             interval="1d"
         )
 
+
+@patch('san.graphql.requests.post')
+def test_get_without_transform(mock):
+    expected = {
```
IvanIvanoff: `api_call_result` (also rename the other places where `expected` is used for the mock)

san/transform.py
spiderjako: Did the comment on 2 lines

san/transform.py
spiderjako: You're right, just did that

san/tests/test_get.py
spiderjako: Just did that, it really does confuse

san/tests/test_get.py
spiderjako: Just finished up the renaming at a few places

PR #76, opened 2019-08-01
Handle empty results in sanNews

spiderjako:
When there are no news, the list, which is returned is empty, that means it can't be indexed, hence the "index out of range" error. That's why I check if it is empty before that

### DIFF
--- a/san/transform.py
+++ b/san/transform.py
@@ -60,9 +60,9 @@ def news_transform(data):
         'url': column['url']
     }, data))
 
-    df = pd.DataFrame(result, columns = result[0].keys())
-
-    return df
+    if result:
+        df = pd.DataFrame(result, columns = result[0].keys())
+        return df
 
 
 def token_top_transactions_transform(data):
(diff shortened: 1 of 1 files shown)


san/transform.py
```diff
@@ -60,9 +60,9 @@ def news_transform(data):
         'url': column['url']
     }, data))
 
-    df = pd.DataFrame(result, columns = result[0].keys())
-
-    return df
+    if result:
+        df = pd.DataFrame(result, columns = result[0].keys())
```
tspenov: comment on the already commited code:
why news_transform returns a dataframe and other _transform a list ?

spiderjako: The news_transform returns this dataframe, because of the result[0].keys(), if this is not the case, then the columns won't be ordered the way they should be. That was necessary for the news_transform function, for the others it isn't, that's why they return lists.

tspenov: I would use `OrderedDict` instead of `{}` in order to achieve sorting and unify result type of _transform functions. Currently it is not a problem cause result is then passed to `convert_to_datetime_idx_df` which doesn't care if it is a dataframe or list but if you add something in the middle the problem will show.

spiderjako: You're right, it works just fine and is more practical this way

PR #77, opened 2019-08-12
Add get metric to sanpy

spiderjako:
Migration of get metric API to Sanpy


san/get.py
```diff
@@ -12,9 +12,12 @@ def get(dataset, **kwargs):
     query, slug = parse_dataset(dataset)
     if query in CUSTOM_QUERIES:
         return getattr(san.sanbase_graphql, query)(0, slug, **kwargs)
-
-    gql_query = get_gql_query(0, dataset, **kwargs)
-    gql_query = "{\n" + gql_query + "\n}"
+    if query in V2_METRIC_QUERIES:
+        gql_query = san.sanbase_graphql.get_metric(0, query, slug, **kwargs)
+        gql_query = "{\n" + gql_query + "\n}"
```
IvanIvanoff: GQL does not care if you pass `{ query}` or `{\n query \n}`. In this case, we're aiming to reduce the request size, so let's just do `gql_query =  "{" + san.sanbase_graphql.get_metric(0, query, slug, **kwargs) + "}"`

san/sanbase_graphql.py
```diff
@@ -464,6 +464,32 @@ def ohlcv(idx, slug, **kwargs):
     return merged[return_fields]
 
 
+def get_metric(idx, query, slug, **kwargs):
```
IvanIvanoff: rename `query` to `metric`

san/sanbase_graphql.py
```diff
@@ -464,6 +464,32 @@ def ohlcv(idx, slug, **kwargs):
...
+    kwargs = _transform_query_args(**kwargs)
+
+    query_str = """
+    query_{idx}: getMetric(metric: \"{query}\"){{
+        timeseriesData(
+            slug: \"{slug}\",
+            from: \"{from_date}\",
+            to: \"{to_date}\",
+            interval: \"{interval}\",
```
IvanIvanoff: GraphQL has insignificant commas, I would prefer we do not add commas where not needed: https://github.com/graphql/graphql-spec/blob/master/spec/Section%202%20--%20Language.md#insignificant-commas

san/transform.py
```diff
@@ -11,9 +11,11 @@
 QUERY_PATH_MAP = {
     'eth_top_transactions': ['ethTopTransactions'],
     'eth_spent_over_time': ['ethSpentOverTime'],
-    'token_top_transactions':['tokenTopTransactions']
+    'token_top_transactions':['tokenTopTransactions'],
+    'get_metric':['timeseriesData']
 }
 
+V2_METRIC_QUERIES = ['average_price', 'daily_active_addresses', 'mvrv_usd_20y']
```
IvanIvanoff: Why these queries only?
`average_price` should be removed, too

IvanIvanoff: And we need to add much more metrics here: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/clickhouse/metric/available_v2_metrics.json

spiderjako: This is temporary, just wanted to see if it should be implemented like this or it should be changed. Will add them in a bit

san/batch.py
```diff
@@ -1,6 +1,7 @@
+import san.sanbase_graphql
 from san.query import get_gql_query
 from san.graphql import execute_gql
-from san.transform import transform_query_result
+from san.transform import transform_query_result, V2_METRIC_QUERIES
```
IvanIvanoff: I do not like having to import `V2_METRIC_QUERIES` from `san.transform` - I would never search for it there in the first place. We can have a module with appropriate name that could hold this list?

san/batch.py
```diff
@@ -16,7 +17,12 @@ def execute(self):
         batched_queries = []
 
         for idx, q in enumerate(self.queries):
-            batched_queries.append(get_gql_query(idx, q[0], **q[1]))
+            metric = q[0].split('/')[0]
+            slug = q[0].split('/')[1]
```
IvanIvanoff:
```python
[metric, slug] = str.split("/")
```
to avoid splitting twice

san/batch.py
```diff
@@ -16,7 +17,12 @@ def execute(self):
         batched_queries = []
 
         for idx, q in enumerate(self.queries):
```
IvanIvanoff: Let's also rework this to be `query` instead of `q`. The name `q` does not mean anything

san/batch.py
IvanIvanoff: Also, what is this `q` exactly so we need to fetch the thing at position `0`?

san/get.py
```diff
@@ -12,9 +12,12 @@ def get(dataset, **kwargs):
...
+    else:
```
IvanIvanoff: Why the first `if` has an early return and this if/else assigns to a variable and uses it? Let's unify

san/sanbase_graphql.py
```diff
@@ -464,6 +464,32 @@ def ohlcv(idx, slug, **kwargs):
...
+    kwargs = _transform_query_args(**kwargs)
+
+    query_str = """
+    query_{idx}: getMetric(metric: \"{metric}\"){{
+        timeseriesData(
+            slug: \"{slug}\"
+            from: \"{from_date}\"
+            to: \"{to_date}\"
+            interval: \"{interval}\"
+            aggregation: {aggregation}
```
IvanIvanoff: this should have `null` as default

san/transform.py
```diff
@@ -11,9 +11,78 @@
...
 
+V2_METRIC_QUERIES = [
+    'daily_active_addresses',
+    'daily_avg_marketcap_usd',
+    'daily_avg_price_usd',
+    'daily_closing_marketcap_usd',
+    'daily_closing_price_usd',
+    'daily_high_price_usd',
+    'daily_low_price_usd',
+    'daily_opening_price_usd',
```
IvanIvanoff: remove all price-related metrics from here

san/batch.py
spiderjako: You're right, I was going to ask you that.

san/batch.py
spiderjako: q[0] is the dataset, and q[1] are the kwargs, it can be seen in the get() method

san/get.py
spiderjako: It is because of the OHLC function

san/transform.py
spiderjako: I guess without realized price, I will do it

san/sanbase_graphql.py
spiderjako: Yes, it should be so

san/batch.py
```diff
@@ -15,8 +16,14 @@ def execute(self):
         result = []
         batched_queries = []
 
-        for idx, q in enumerate(self.queries):
-            batched_queries.append(get_gql_query(idx, q[0], **q[1]))
+        for idx, query in enumerate(self.queries):
```
tspenov: This function is getting very complex. We can split it into smaller named functions.
lets have extract: 
```python
        batched_queries = []

        for idx, query in enumerate(self.queries):
            [metric, slug] = query[0].split('/')
            if metric in V2_METRIC_QUERIES:
                batched_queries.append(
                    san.sanbase_graphql.get_metric(
                        idx, metric, slug, **query[1]))
            else:
                batched_queries.append(get_gql_query(idx, query[0], **query[1]))

        gql_string = self.__batch_gql_queries(batched_queries)
```
into `__create_batched_query_string`.

So we end up with:
```
def execute(self):
    graphql_string = self.__create_batched_query_string()
    result = execute_gql(gql_string)
    return self.__transform_batch_result(result)
```

spiderjako: It will be truly more readable, I will try and make it so if you wish. @IvanIvanoff  @kalkov What do you think?

san/batch.py
```diff
@@ -11,24 +12,36 @@ def __init__(self):
...
+        idxs = sorted([int(k.split('_')[1]) for k in executed_gql.keys()])
         for idx in idxs:
             query = self.queries[idx][0].split("/")[0]
-            df = transform_query_result(idx, query, res)
+            df = transform_query_result(idx, query, executed_gql)
             result.append(df)
 
         return result
 
+    def execute(self):
```
tspenov: lets move this below `get` because that is our public interface. And move helper functions starting with `__` to the bottom

spiderjako: Oh, my bad, I will do it

san/batch.py
```diff
@@ -11,24 +12,36 @@ def __init__(self):
...
+            if metric in V2_METRIC_QUERIES:
+                batched_queries.append(
+                    san.sanbase_graphql.get_metric(
+                        idx, metric, slug, **query[1]))
+            else:
+                batched_queries.append(get_gql_query(idx, query[0], **query[1]))
+        self.__batch_gql_queries(batched_queries)
+        return self.__batch_gql_queries(batched_queries)
+
+    def __transform_batch_result(self, executed_gql):
```
tspenov: `executed_gql` -> `graphql_result` :?

spiderjako: It does sound better, change is on the way

tspenov:
@spiderjako Good work on this one 👏 . I think what is left is to add example in the README, link to where a list of all metrics can be seen and bump the version ti `0.7.0` because it is a major change.

README.md
```diff
@@ -1195,6 +1196,32 @@ datetime                           fromAddress  fromAddressInExchange
 2019-04-30 15:17:28+00:00  0x876eabf441b2ee...                   True  0x1f4a90043cf2d...                False  0xc85892b9ef8c64...   20544.42975
 ```
 
+### Get Metric
```
tspenov: @IvanIvanoff @kalkov Do you have Idea how to better present these metrics in the README. Now it looks to me like its just another metric but its a pretty big feature.

san/v2_metrics_list.py
```diff
@@ -0,0 +1,62 @@
+V2_METRIC_QUERIES = [
```
IvanIvanoff: Can we replace this by an API call to https://api.santiment.net/graphiql?variables=&query=%7B%0A%20%20getAvailableMetrics%0A%7D%0A?

spiderjako: Yes it can, I will only implement the getAvailableMetrics and call it, but is it better than just having the file? Another thing, the README points exactly to that file, where the available metrics can be seen.

IvanIvanoff: The current list will be outdated when I merge https://github.com/santiment/sanbase2/pull/1398

We can have README data that is not connected to the list here :?

spiderjako: You're right, that it will be outdated.

README.md
kalkov: @tspenov For sure we need to have a list of all metrics here(first function argument)

san/v2_metrics_list.py
```diff
@@ -0,0 +1,63 @@
...
+    'stack_realized_cap_usd_365d',
+    'stack_realized_cap_usd_3y',
+    'stack_realized_cap_usd_5y',
+    'stack_realized_cap_usd_60d',
+    'stack_realized_cap_usd_7d',
+    'stack_realized_cap_usd_90d',
+    'stack_total_age',
+    'token_velocity',
+    'transaction_volume'
+]
```
IvanIvanoff: This list is outdated, this is the current list: https://api.santiment.net/graphiql?variables=&query=%7B%0A%20%20getAvailableMetrics%0A%7D%0A

spiderjako: It should be okay now

V2_METRICS.md
```diff
@@ -0,0 +1,74 @@
+# Get Metric
+
+Returns data for a given metric. The input, that is needed, is the string 'metric/slug', a 'from' date, a 'to' date, a string for the interval and an aggregation, which is optional (It is 'NULL' by default).
```
tspenov: I think only the available metrics should stay in this file. If you want to explain the options it should be in the README.md before the example. Also `aggregation` - when it it not passed as argument it uses the aggregation determined from the API.

requirements.txt
```diff
@@ -0,0 +1,5 @@
+pandas
```
tspenov: I think adding requirements.txt without any options or pinned versions does nothing different than using `install_requires`: https://github.com/santiment/sanpy/blob/master/setup.py#L20

V2_METRICS.md
spiderjako: You're right, done as you've said

requirements.txt
spiderjako: It is a bit noisy, so I've removed it.

README.md
```diff
@@ -1195,6 +1196,31 @@ datetime                           fromAddress  fromAddressInExchange
...
```
IvanIvanoff: having `Get Metric` in the documentation is an implementation detail that could be omitted

V2_METRICS.md
```diff
@@ -0,0 +1,72 @@
...
+      "realized_value_usd_5y"
+      "realized_value_usd_3y"
+      "realized_value_usd_2y"
+      "realized_value_usd_365d"
+      "realized_value_usd_180d"
+      "realized_value_usd_90d"
+      "realized_value_usd_60d"
+      "realized_value_usd_30d"
+      "realized_value_usd_7d"
+      "realized_value_usd_1d"
```
IvanIvanoff: Add some docs what these `_1d`, `_7d` etc. mean

README.md
spiderjako: Why wouldn't it be in the documentation?

PR #78, opened 2019-09-05
Restructuring of the default query

spiderjako:


san/sanbase_graphql.py
```diff
@@ -106,6 +105,20 @@
     'projects': {
         'query': 'allProjects',
         'return_fields': ['name', 'slug', 'ticker', 'totalSupply', 'marketSegment']
+    },
+    'eth_top_transactions': {
+        'query': 'ethTopTransactions',
+        'return_fields': [
+            'datetime',
+            {
+                'fromAddress': ['address', 'isExchange']
```
IvanIvanoff: This should always be a map with a single key-value pair. In this case, this could be a tuple instead?
```python
'return_fields': [
'datetime',
('fromAddress', ['address', 'isExchange']),
...
```

san/sanbase_graphql.py
```diff
@@ -322,29 +335,23 @@ def price_volume_difference(idx, slug, **kwargs):
 
 def eth_top_transactions(idx, slug, **kwargs):
     kwargs = _transform_query_args('eth_top_transactions', **kwargs)
+    return_fields = []
+    for el in kwargs['return_fields']:
+        if isinstance(el,dict):
+            el = str(
+                list(el.keys())[0]
```
IvanIvanoff: this weird `[0]` (why ignore all other keys someone without context knowledge my ask) can be avoided by the above suggested dict -> tuple migration

san/sanbase_graphql.py
```diff
@@ -98,6 +97,24 @@
...
         'query': 'historyTwitterData',
         'return_fields': ['datetime', 'followers_count']
+    },
+    'historical_balance': {
+        'query': 'historicalBalance',
+        'return_fields': ['datetime', 'balance']
+    },
+    'projects': {
+        'query': 'allProjects',
+        'return_fields': ['name', 'slug', 'ticker', 'totalSupply', 'marketSegment']
```
IvanIvanoff: let's do a follow-up PR where we extend the list of the default returned fields

san/sanbase_graphql.py
```diff
@@ -228,20 +245,17 @@ def mining_pools_distribution(idx, slug, **kwargs):
...
         slug: \"{slug}\",
         from: \"{from_date}\",
         to: \"{to_date}\",
         interval: \"{interval}\"
     ){{
-        balance,
-        datetime
-    }}
-    """.format(
+    """ + ' '.join(kwargs['return_fields']) + '}}').format(
```
IvanIvanoff: How does `join` know that it needs to convert the tuple  ` ('fromAddress', ['address', 'isExchange'])` to `fromAddress{address isExchange}` :?

spiderjako: Hmm, good question. https://github.com/santiment/sanpy/pull/78/files/dfd80d394fc15a935a3fdc79a3f32e551c7a49e6#diff-0f5457262476aa06c3885256259a709dR335
I will do something similar here, thanks for the heads-up

IvanIvanoff: It works (I have tested it locally) - I just do not know why

spiderjako: Now it should know

san/sanbase_graphql.py
```diff
@@ -228,20 +245,17 @@ def mining_pools_distribution(idx, slug, **kwargs):
 
 
 def historical_balance(idx, slug, **kwargs):
-    kwargs = _transform_query_args(**kwargs)
+    kwargs = _transform_query_args("historical_balance", **kwargs)
```
IvanIvanoff: consistently use `"` and `'`

spiderjako: Hm, my bad, it should be with '_'

san/sanbase_graphql.py
```diff
@@ -611,6 +577,15 @@ def _transform_query_args(**kwargs):
     kwargs['from_date'] = _format_from_date(kwargs['from_date'])
     kwargs['to_date'] = _format_to_date(kwargs['to_date'])
 
+    return_fields = []
+    if 'return_fields' in kwargs:
+        _format_return_fields(return_fields, kwargs['return_fields'])
+        kwargs['return_fields'] = return_fields
```
tspenov: We can we simplify like this: `kwargs['return_fields'] = format_return_fields(kwargs['return_fields'])`

san/sanbase_graphql.py
```diff
@@ -640,3 +615,10 @@ def _format_to_date(datetime_obj_or_str):
     dt = iso8601.parse_date(datetime_obj_or_str) + \
         datetime.timedelta(hours=23, minutes=59, seconds=59)
     return dt.isoformat()
+
+def _format_return_fields(list_to_be_modified, list_to_be_iterated):
```
tspenov: we can use `map` - this eliminates need for `list_to_be_modified`

spiderjako: You're right, it does optimise things a bit

san/sanbase_graphql.py
```diff
@@ -640,3 +612,8 @@ def _format_to_date(datetime_obj_or_str):
     dt = iso8601.parse_date(datetime_obj_or_str) + \
         datetime.timedelta(hours=23, minutes=59, seconds=59)
     return dt.isoformat()
+
+def _format_return_fields(list_to_be_iterated):
```
tspenov: `list_to_be_iterated` can be renamed to `fields` cause there is no more need to distinguish the two params cause  now we have only one.

san/sanbase_graphql.py
```diff
@@ -640,3 +667,8 @@ def _format_to_date(datetime_obj_or_str):
     dt = iso8601.parse_date(datetime_obj_or_str) + \
         datetime.timedelta(hours=23, minutes=59, seconds=59)
     return dt.isoformat()
+
+def _format_return_fields(fields):
+    return list(map(
+        lambda el: el[0] + '{{' + ' '.join(el[1]) + '}}' if isinstance(el, tuple) else el
```
IvanIvanoff: it's possible to have deeply nested fields - this won't handle them properly. what happens if you have a tuple that contains a tuple?

spiderjako: You're right, that's a good question

spiderjako: Should be okay now

san/batch.py
```diff
@@ -21,11 +22,12 @@ def __create_batched_query_string(self):
         batched_queries = []
 
         for idx, query in enumerate(self.queries):
-            [metric, slug] = query[0].split('/')
+            [metric, _separator, slug] = query[0].partition('/')
```
IvanIvanoff: why change to `partition` when you ignore the separator?

san/batch.py
```diff
@@ -21,11 +22,12 @@ def __create_batched_query_string(self):
...
             if metric in V2_METRIC_QUERIES:
-                batched_queries.append(
-                    san.sanbase_graphql.get_metric(
-                        idx, metric, slug, **query[1]))
+                if slug != '':
+                    batched_queries.append(san.sanbase_graphql.get_metric(idx, metric, slug, **query[1]))
+                else:
+                    raise SanError('Invalid metric!')
             else:
```
IvanIvanoff: why we check if slug is not empty only in the case of v2 metric?

san/tests/test_sanbase_integration.py
```diff
@@ -13,10 +13,33 @@
     "address": "0x1f3df0b8390bb8e9e322972c5e75583e87608ec2"
 }
 
+METRICS_EQUAL_FORMAT = [
```
IvanIvanoff: add a comment what does this mean

san/batch.py
spiderjako: We check it also in get_gql_query, the checking is just in the san.query

san/batch.py
spiderjako: Because of split's nature - when we do [metric, slug] = ...split('/'), and there isn't a '/', then there isn't a value for slug, whereas partition just makes slug = ''

san/tests/test_sanbase_integration.py
spiderjako: Alright

san/sanbase_graphql.py
```diff
@@ -691,3 +709,15 @@ def _format_to_date(datetime_obj_or_str):
     dt = iso8601.parse_date(datetime_obj_or_str) + \
         datetime.timedelta(hours=23, minutes=59, seconds=59)
     return dt.isoformat()
+
+def _format_all_return_fields(fields):
+    while any(isinstance(x, tuple) for x in fields):
+        fields = list(map(
```
tspenov: Shouldn't you use `_format_return_fields(fields)` here ?

san/sanbase_graphql.py
```diff
@@ -691,3 +709,15 @@ def _format_to_date(datetime_obj_or_str):
...
+        lambda el: el[0] + '{{' + ' '.join(el[1]) + '}}' if isinstance(el, tuple) else el
+        , fields))
+    return fields
+
+def _format_return_fields(fields):
```
tspenov: This function is defined twice in this file and not used

PR #79, opened 2019-09-09
Add trending-words

spiderjako:


san/query.py
```diff
@@ -3,13 +3,19 @@
...
 
 def get_gql_query(idx, dataset, **kwargs):
-    query, slug = parse_dataset(dataset)
-    return getattr(
-        san.sanbase_graphql,
-        query,
-        lambda *args, **kwargs: not_found(query)
-    )(idx, slug, **kwargs)
-
+    if "/" in dataset:
```
IvanIvanoff: What's `dataset` here? It seems to me that it's a string. 
Also, what's the logic behind `if "/" in dataset`?

spiderjako: The dataset is a string, containing "metric/slug", so if it doesn't have a "/" there, then there isn't a slug. The problem is, that one of the tests is failing atm and I'm trying to figure that one out

san/sanbase_graphql.py
```diff
@@ -440,6 +440,28 @@ def token_top_transactions(idx, slug, **kwargs):
...
+    query_str = """
+    query_{idx}: getTrendingWords (
+        from: \"{from_date}\",
+        to: \"{to_date}\",
+        size: {size},
+        interval: \"{interval}\"
+    ){{
+        datetime,
+        topWords{{
+            score,
```
IvanIvanoff: omit the commas in the query

san/query.py
```diff
@@ -2,15 +2,21 @@
 from san.error import SanError
 
 
-def get_gql_query(idx, dataset, **kwargs):
-    query, slug = parse_dataset(dataset)
+def get_gql_query(idx, identifier, **kwargs):
+  query, _separator, slug = identifier.partition("/")
+  if slug == '':
```
IvanIvanoff: btw, this is not guarding against passing something like `daily_active_addresses/`. Maybe the first case should be `if separator != '' and slug == '': <error>`

README.md
```diff
@@ -1233,6 +1234,36 @@ datetime                      value
 2019-09-03 00:00:00+00:00    9.0
 ```
 
+### Trending Words
```
IvanIvanoff: Let's use the new name `Emerging Trends` instead of `Trending Words` even though the API has the old name

PR #81, opened 2019-10-07
Add available metrics function

spiderjako:


san/sanbase_graphql.py
```diff
@@ -193,7 +193,10 @@
             'datetime',
             ('topWords', ['score', 'word'])
         ]
-    }     
+    },
+    'social_projects': {
+        'return_fields':[]
```
tspenov: what is social_projects and why it has return fields `[]` and why in this PR ?

tspenov: is it equivalent to `socialVolumeProjects` from api ?

spiderjako: Because that's the only metric in sanbase_graphql.py, which doesn't have an instance in QUERY_MAPPINGS, because it never uses the transform_query_args function. Also, we do need ALL of the functions in sanpy, that's why I added the function in QUERY_MAPPINGS.

spiderjako: Maybe I should remove the 'return_fields', I had just added that for consistency

README.md
```diff
@@ -1278,6 +1278,14 @@ datetime                              slug     change    status
 2019-07-30 11:00:00+00:00            litex   8.000000  NEWCOMER
 ```
 
+### Available metrics
+
+Gets ahold of a list with all the metrics in sanpy.
```
tspenov: this should be put here:
image
instead of the manual bullet list list
Also correct the link from Table of contents - cause it doesn't go here

tspenov:
Also bump the version

san/sanbase_graphql.py
tspenov: do we have an example for this in the README ? I see `social_volume_projects` there

IvanIvanoff:
```
Because that's the only metric in sanbase_graphql.py, which doesn't have an instance in QUERY_MAPPINGS, because it never uses the transform_query_args function
```
then let's make it use it so it's not handled in this weird inconsitent way

san/sanbase_graphql.py
```diff
@@ -1,6 +1,6 @@
 import iso8601
 import datetime
-from san.pandas_utils import merge
```
spiderjako: It had to be done, so that inspect.getmembers() doesn't count 'merge' as a function from this file.

san/available_metrics.py
```diff
@@ -1,4 +1,12 @@
+import inspect
 import san.sanbase_graphql
 
 def available_metrics():
-    return san.sanbase_graphql.available_metrics()
+    all_functions = inspect.getmembers(san.sanbase_graphql, inspect.isfunction)
+    all_functions = list(map(
+        lambda x: x[0] if x[0][0] != '_' else all_functions.remove(x), all_functions
```
IvanIvanoff: What is this?

san/available_metrics.py
```diff
@@ -0,0 +1,12 @@
...
+import san.sanbase_graphql
+
+def available_metrics():
+    all_functions = inspect.getmembers(san.sanbase_graphql, inspect.isfunction)
+    all_functions = list(map(
+        lambda x: x[0] if x[0][0] != '_' else all_functions.remove(x), all_functions
+    ))
+    return list(filter(
+        lambda x: x != None, all_functions
+    ))
```
IvanIvanoff:
```python
return list(filter(None, all_functions)) 
```

san/available_metrics.py
IvanIvanoff: Also, to every element, you're mapping either the whole list or the whole list with 1 element removed?
I guess it's holding a reference and removing something at iteration N affects the element at position N - 1? 
Either way, please rewrite this

spiderjako: You're right, should be okay now

san/sanbase_graphql_helper.py
```diff
@@ -0,0 +1,268 @@
+import iso8601
+import datetime
+
+_DEFAULT_INTERVAL = '1d'
```
IvanIvanoff: You no longer need to prepend `_` to functions and vars in this module

spiderjako: I do, because when I import in sanbase_graphql from sanbase_graphql_helper, DEFAULT_INTERVAL gets imported needlessly, hence the '_'.

IvanIvanoff: if it starts with `_` it's not imported?

spiderjako: Yes, exactly

IvanIvanoff: Then why you needed to remove the functions starting with `_` in the previous implementation?

spiderjako: Because inspect.getmember returns all of the functions (there isn't any real encapsulation in python), but when you import stuff from another module, python is made so, that you cannot import functions starting with '_' or '__'

IvanIvanoff:
btw, how can users know what arguments to pass to a function once they know the name?

PR #83, opened 2019-10-10
Feature/httpx

blockchainunchained:
Add httpx and then use it to support async clients.

### DIFF
--- a/san/__init__.py
+++ b/san/__init__.py
@@ -3,6 +3,7 @@ from .metadata import metadata
 from .available_metrics import available_metrics
 from .batch import Batch
 from .api_config import ApiConfig
+import asyncio
 import pkg_resources
 import httpx
 import json
@@ -11,18 +12,19 @@ from warnings import warn
 PROJECT = 'sanpy'
 
 
-def get_latest():
+async def get_latest():
     url = 'https://pypi.python.org/pypi/%s/json' % (PROJECT)
     try:
-        response = httpx.get(url).text
-        return json.loads(response)['info']['version']
+        async with httpx.AsyncClient() as client:
+            response = await client.get(url)
+        return json.loads(response.text)['info']['version']
     except httpx.exceptions.HTTPError as e:
         return pkg_resources.get_distribution(PROJECT).version
 
 
-def warn_if_outdated():
+async def warn_if_outdated():
     current_version = pkg_resources.get_distribution(PROJECT).version
-    latest_version = get_latest()
+    latest_version = await get_latest()
 
     if current_version != latest_version:
         warning = 'The package %s is out of date. Your version is %s, the latest is %s.' % (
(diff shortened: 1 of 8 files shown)


IvanIvanoff:
image

I think we should proceed with moving this to production once httpx is stable

blockchainunchained:
Agreed, I'm using this branch to power the ticker at /r/ethfinance and it's stable for my use but I'm not doing anything crazy or weird.

valo:
@blockchainunchained is there any change on the public API with this case? If not, what are the benefits from this change? If yes, should we add some more info in the README?

@spiderjako can you, please, take a look at this PR and see if you have some comments?

blockchainunchained:
No changes to the public API, this simply uses HTTPX (who will be releasing 1.0 shortly, API should already be stable) instead of Requests. HTTPX is designed to be a drop in replacement for requests but allows asynchronous projects to not block their main thread when trying to interact with SanBase via SanPy.

I've not pulled the latest HTTPX or the latest SanPy recently so don't merge this until I've had a chance to test it (and also we should wait until HTTPX goes 1.0).

No documentation updates are required because if an event loop is not detected one is created internally so it doesn't matter whether you are running synchronous Python code or asynchronous Python code, or at least that is my understanding of it anyway.

kalkov:
@blockchainunchained, @valo  
Maybe would be best to close the PR now and re-open when HTTPX release 1.0?

blockchainunchained:
I've been keeping this up to date with the latest HTTPX and I'm more than happy to continue doing so until the final version of HTTPX is released. Other than artificially lowering the pull request count what is the advantage to closing this pull request prematurely?

IvanIvanoff:
You can just put a `WIP` label on it and leave it open

blockchainunchained:
It doesn't look like labels are setup, do you mean edit the title, if so should it prefix or postfix the existing title?

IvanIvanoff:
I've created this label

IvanIvanoff:
`httpx` still has no 1.0 version ~5 years later.

Over the years since this PR was opened we've did some work in the async direction:
- https://github.com/santiment/sanpy/pull/151
- https://github.com/santiment/sanpy/pull/155

After some discussion internally, we decided that we don't want to replace the internals as done in this PR. One of our experienced Python developers has the following concerns:
- Making base implementation async, and then using try/catch for sync versions exposes ordinary users to weird event loop interactions
- There is no need of making all calls async, e.g. warn_if_outdated is used in module import and we don't get any benefits from it
- `httpx` still hasn't reached stable release (latest version is 0.27.0)
- In case our library calls are required to be async, we propose a `ThreadPoolExecutor` + `loop.run_in_executor` solution to wrap our execute calls to unblock main thread, while `sanpy` fetches data


If we still find the need to do something similar to the PR, we'll do it as an alternative implementation to the existing one, and users of the library can choose what to do.

PR #85, opened 2019-10-19
Upgrade Backtest class

alexander-gr:
Fix bugs in the Backtest class.
Remove unused code and improve syntax.
Add visualization method.

### DIFF
--- a/san/backtest.py
+++ b/san/backtest.py
@@ -2,53 +2,44 @@ import datetime
 import pandas as pd
 import numpy as np
 import san
+import matplotlib.pyplot as plt
 
 class Backtest:
 
-    def __init__(self, returns, trades, lagged=True, transaction_cost = 0, percent_invested_per_trade = 1):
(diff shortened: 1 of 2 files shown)


san/backtest.py
```diff
@@ -2,53 +2,44 @@
 import pandas as pd
 import numpy as np
 import san
+import matplotlib.pyplot as plt
```
tspenov: Should we put `matplotlib` in https://github.com/santiment/sanpy/blob/master/setup.py#L20 ?

tspenov:
@alexander-gr Maybe we should add some usage info in the README ?

tspenov:
@alexander-gr You can also bump the version, so we can release it

san/backtest.py
alexander-gr: Yep, sure
Thanks!

alexander-gr:
> @alexander-gr Maybe we should add some usage info in the README ?

As far as I know Jan is working on all docs ATM

PR #87, opened 2019-11-04
Optimizing backtest

alexander-gr:
Optimized perfomance of Backtest class.
Fixed bug with maxDrawdown.
New viz for provided trades.
Prettify the code.

### DIFF
--- a/san/backtest.py
+++ b/san/backtest.py
@@ -4,14 +4,14 @@ import numpy as np
 import san
 import matplotlib.pyplot as plt
 
+
 class Backtest:
 
     def __init__(self, returns:pd.Series, trades:pd.Series, lagged=True, transaction_cost = 0, percent_invested_per_trade = 1):
         """ Initializing Backtesting function
-            Init function generates performance of the test and several risk metrics. The object lets
-            you specify wether you want to lag the trades to avoid overfitting, the transaction costs
+            Init function generates performance of the test. The object lets you specify
+            wether you want to lag the trades to avoid overfitting, the transaction costs
             and the percentage of the portfolio to be invested per trade (50% as 0.5).
-
             Trade example:
                 With given prices [P1, P2, P3] and given trades [False, True, False] 
                 you buy asset when the price is P2 and sell when the price is P3.
@@ -22,7 +22,6 @@ class Backtest:
             trades.iloc[0] = False
         self.strategy_returns = ((returns * percent_invested_per_trade) * trades)
         self.trades = trades
-        
         self.nr_trades = 0
         for i in range(1, len(trades)):
             if trades[i] != trades[i - 1]:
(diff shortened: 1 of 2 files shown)


san/backtest.py
```diff
@@ -22,40 +22,40 @@ def __init__(self, returns:pd.Series, trades:pd.Series, lagged=True, transaction
             trades.iloc[0] = False
         self.strategy_returns = ((returns * percent_invested_per_trade) * trades)
         self.trades = trades
-        
-        self.nr_trades = 0
+
+        self.nr_trades = {'buy': [], 'sell': []}
```
IvanIvanoff: `nr` was coming from `number`? if yes - this no longer keeps the number of trades, but list of trades

san/backtest.py
```diff
@@ -22,40 +22,40 @@ def __init__(self, returns:pd.Series, trades:pd.Series, lagged=True, transaction
...
-        a = np.argmax(np.maximum.accumulate(running_value) - running_value)  # end of the period
-        b = np.argmax(running_value[:a])  # start of period
-        self.maximum_drawdown = ((running_value[a] - running_value[b]) / running_value[b]) * 100  # Maximum Drawdown
-
-
-    def get_sharpe_ratio(self):
-        return round(self.sharpe_ratio, 2)
 
+    def get_sharpe_ratio(self, decimals=2):
+        sharpe_ratio = (self.strategy_returns.mean() * 365) / (self.strategy_returns.std() * np.sqrt(365))
```
IvanIvanoff: should this always be 365 or it should be configurable?

san/backtest.py
```diff
@@ -22,40 +22,40 @@ def __init__(self, returns:pd.Series, trades:pd.Series, lagged=True, transaction
...
 
     def get_nr_trades(self):
-        return self.nr_trades
+        return len(self.nr_trades['sell']) + len(self.nr_trades['buy'])
 
-    def get_maximum_drawdown(self):
-        return round(self.maximum_drawdown, 2)
+    def get_maximum_drawdown(self, decimals=2):
+        running_value = np.array(self.performance)
+        running_value[0] = 0
```
IvanIvanoff: I guess you're right, but why the first element is made zero

san/backtest.py
JanSmirny: Its annualizing the sharpe ratio. The issue with this is that it only works for daily data of course. We could try some mechanism that figures out the granularity and then annualizes accordingly. I think pandas has something for this. But I guess this would also make it more fragile since the entered dataframe needs precisely specified datetime...

san/backtest.py
```diff
@@ -22,40 +22,40 @@ def __init__(self, returns:pd.Series, trades:pd.Series, lagged=True, transaction
...
+        end = np.argmax(np.maximum.accumulate(running_value) - running_value)  # end of the dropdown period
+        start = np.argmax(running_value[:end])  # start of the dropdown period
+        maximum_drawdown = running_value[start] - running_value[end]
```
JanSmirny: Should we try `(end - start) / start` here instead to get the maximal drop in percent?

san/backtest.py
alexander-gr: yep
but variable trades already taken and nr_trades is mostly used to count number of sell/buy trades

san/backtest.py
alexander-gr: thats because we shift the trades. So its nan in the first day (or just 0). You dont sell in the first day so returns = 0

san/backtest.py
alexander-gr: we didnt go to any conclusion yet but I really need updated sanpy tomorrow. I propose leave it as it is now and open a new PR for that. (so I will not use it tomorrow)

PR #88, opened 2019-11-18
Changing the description of sanpy

spiderjako:
When you input:

from_date = '2018-01-01'
to_date = '2019-11-01'
asset = 'ethereum'
interval = '1d'

san.get( f'daily_active_addresses/{asset}', from_date=from_date, to_date=to_date, interval=interval )

you get a data frame with a column with the name of 'value' instead of 'daily active addresses'

I've fixed the description in the README and removed the functions, which were for the old API.

### DIFF
--- a/san/get.py
+++ b/san/get.py
@@ -9,9 +9,20 @@ CUSTOM_QUERIES = {
     'ohlcv': 'get_ohlcv'
 }
 
+DEPRECATED_QUERIES = {
+    'mvrv_ratio': 'mvrv_usd',
+    'nvt_ratio': 'nvt',
+    'realized_value': 'realized_value_usd',
+    'token_circulation': 'circulation_1d',
+    'burn_rate': 'age_destroyed',
+    'token_age_consumed': 'age_destroyed',
+    'token_velocity': 'velocity'
+}
 
 def get(dataset, **kwargs):
     query, slug = parse_dataset(dataset)
+    if query in DEPRECATED_QUERIES:
+        print(f'**NOTICE**\n{query} will be deprecated in version 0.9.0, please use {DEPRECATED_QUERIES[query]} instead')
     if query in CUSTOM_QUERIES:
         return getattr(san.sanbase_graphql, query)(0, slug, **kwargs)
     if query in V2_METRIC_QUERIES:
--- a/san/sanbase_graphql.py
+++ b/san/sanbase_graphql.py
@@ -3,23 +3,12 @@ import san.sanbase_graphql_helper as sgh
 from san.batch import Batch
 from san.error import SanError
 
-def daily_active_addresses(idx, slug, **kwargs):
-    query_str = sgh.create_query_str(
-        'daily_active_addresses', idx, slug, **kwargs)
-
-    return query_str
 
 # to be removed
 
 
 def burn_rate(idx, slug, **kwargs):
     query_str = sgh.create_query_str('burn_rate', idx, slug, **kwargs)
-
-    return query_str
-
-
-def transaction_volume(idx, slug, **kwargs):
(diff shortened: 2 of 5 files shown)


san/sanbase_graphql.py
```diff
@@ -18,12 +13,6 @@ def burn_rate(idx, slug, **kwargs):
     return query_str
 
 
-def transaction_volume(idx, slug, **kwargs):
```
IvanIvanoff: only those 2 should be deleted? They are the ones with the 1:1 names, but mvrv, nvt, circulation, burn rate, etc. are also present in getMetric

IvanIvanoff: Maybe in another PR we need to just point the old names to the new names?

spiderjako: Yeah, I was thinking of doing a PR, in which we just move from the old functions to the new ones, or just point them to the new functions, as you said.

spiderjako: I was thinking, that this PR is for the 1:1 functions and the description change.

spiderjako: I'll note the other PR idea, which we've just discussed.

san/get.py
```diff
@@ -9,9 +9,17 @@
     'ohlcv': 'get_ohlcv'
 }
 
+DEPRECATED_QUERIES_WARNING = [
```
IvanIvanoff: remove `_WARNING`

san/get.py
```diff
@@ -9,9 +9,17 @@
...
+    'mvrv_ratio',
+    'nvt_ratio',
+    'realized_value',
+    'token_circulation'
+]
 
 def get(dataset, **kwargs):
     query, slug = parse_dataset(dataset)
+    if query in DEPRECATED_QUERIES_WARNING:
+        print('**NOTICE**\nThis metric will be deprecated in version 0.9.0, please use the get_metric equivalent instead')
```
IvanIvanoff: If the user executes queries in a loop he'll randomly see `This metric`. Add the metric name.

This warning is not understandable. The users of sanpy have zero knowledge of what `get_metric` is. The warning here should be ` ... mvrv_ratio will be deprecated ... please switch to `mvrv_usd` ... `

spiderjako: Yeah, you're right, somehow mixed the things up, when I was writing them :D I will fix it in a bit

san/get.py
```diff
@@ -9,9 +9,17 @@
     'ohlcv': 'get_ohlcv'
 }
 
+DEPRECATED_QUERIES = {
+    'mvrv_ratio': 'mvrv_usd',
+    'nvt_ratio': 'nvt',
+    'realized_value': 'realized_value_usd',
+    'token_circulation': 'circulation'
+}
```
IvanIvanoff: There are missing ones like: `burn_rate/token_age_consumed -> age_destroyed`

spiderjako: Didn't know about that one, fix is on the way

san/get.py
```diff
@@ -9,9 +9,20 @@
     'ohlcv': 'get_ohlcv'
 }
 
+DEPRECATED_QUERIES = {
+    'mvrv_ratio': 'mvrv_usd',
+    'nvt_ratio': 'nvt',
+    'realized_value': 'realized_value_usd',
+    'token_circulation': 'circulation',
```
IvanIvanoff: the replacement should be `circulation_1d` I think

spiderjako: You're right, thanks

spiderjako: Is it the same with realized_value and realized_value_usd/_1d ?

tspenov:
shouldn't we bump the version ?

PR #89, opened 2019-11-27
Updating the list of metrics with min interval of 5 mins

spiderjako:
age_destroyed_5min and transaction_volume_5min are no longer available with the same names. Also, adding to the list of metrics with min interval 5 min new ones.

### DIFF
--- a/on-chain-metrics.md
+++ b/on-chain-metrics.md
@@ -103,7 +103,10 @@ List of with metrics with 1 day minimal resolution:
 
 - nvt
 
-List of with metrics with 5 minutes minimal resolution:
+List of metrics with 5 minutes minimal resolution:
 
-- transaction_volume_5min
-- age_destroyed_5min
+- transaction_volume
+- age_destroyed
+- exchange_inflow
+- exchange_outflow
+- exchange_balance
--- a/san/tests/test_sanbase_integration.py
+++ b/san/tests/test_sanbase_integration.py
@@ -537,12 +537,10 @@ def test_available_metrics():
         'realized_value_usd_1d',
         'velocity',
         'transaction_volume',
-        'transaction_volume_5min',
         'exchange_inflow',
         'exchange_outflow',
         'exchange_balance',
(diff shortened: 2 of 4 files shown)


san/tests/test_sanbase_integration.py
```diff
@@ -443,7 +443,6 @@ def test_available_metrics():
     functions_list = [
         'average_token_age_consumed_in_days',
         'burn_rate',
-        'daily_active_addresses',
```
IvanIvanoff: ?

san/tests/test_sanbase_integration.py
```diff
@@ -477,7 +476,6 @@ def test_available_metrics():
         'top_holders_percent_of_total_supply',
         'top_social_gainers_losers',
         'topic_search',
-        'transaction_volume',
```
IvanIvanoff: ?

on-chain-metrics.md
```diff
@@ -102,8 +102,3 @@ List of with metrics with 1 day minimal resolution:
 ### NVT
 
 - nvt
-
-List of with metrics with 5 minutes minimal resolution:
-
-- transaction_volume_5min
-- age_destroyed_5min
```
IvanIvanoff: `age_destroyed` and `transaction_volume` are available per 5 minute, only not with this `_5min` name

san/tests/test_sanbase_integration.py
spiderjako: These metrics are no longer available, they're only available from the on-chain list. I thought I've removed them with the last PR, but it turned out, I haven't.

san/tests/test_sanbase_integration.py
spiderjako: Same as the above comment

on-chain-metrics.md
spiderjako: Got it

IvanIvanoff:
bump the version

PR #94, opened 2019-12-14
Fix string formatting for backward compatibility with python < 3.6

spiderjako:
* Fixed the formatting in sanpy to fit every 3.x Python version
* Fixed the graphql handling

### DIFF
--- a/san/graphql.py
+++ b/san/graphql.py
@@ -5,45 +5,47 @@ from san.error import SanError
 
 
 def execute_gql(gql_query_str):
-  headers = {}
-  if ApiConfig.api_key:
-    headers = {'authorization': "Apikey {}".format(ApiConfig.api_key)}
+    headers = {}
+    if ApiConfig.api_key:
+        headers = {'authorization': "Apikey {}".format(ApiConfig.api_key)}
 
-  response = requests.post(
-    SANBASE_GQL_HOST,
-    json={
-        'query': gql_query_str},
-    headers=headers)
+    response = requests.post(
+        SANBASE_GQL_HOST,
+        json={
+            'query': gql_query_str},
+        headers=headers)
 
-  if response.status_code == 200:
-    return __handle_success_response__(response, gql_query_str)
-  else:
-    raise SanError(
-      "Error running query. Status code: {}.\n {}".format(
-        response.status_code,
-        gql_query_str))
+    if response.status_code == 200:
+        return __handle_success_response__(response, gql_query_str)
+    else:
+        raise SanError(
+            "Error running query. Status code: {}.\n {}".format(
+                response.status_code,
+                gql_query_str))
+
 
 def __handle_success_response__(response, gql_query_str):
-  if __result_has_gql_errors__(response):
-    raise SanError(
-      "GraphQL error occured running query {} \n errors: {}".format(
-        gql_query_str,
-        response.json()['errors']))
-  elif __exist_not_empty_result(response):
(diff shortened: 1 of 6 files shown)


san/graphql.py
```diff
@@ -5,45 +5,47 @@
...
+def __handle_graphql_error_response(response, query):
+    if 'errors' in response.json().keys():
+        raise SanError(
+            "GraphQL error occured running query {} \n errors: {}".format(
+                query,
+                response.json()['errors']))
+    else:
+        raise SanError(
+            "Error running query. query: {}\n" .format(
+                query))
```
IvanIvanoff: the status code here is lost - we should include it. If there is an

san/graphql.py
```diff
@@ -5,45 +5,47 @@
...
-  else:
-    raise SanError(
-      "Error running query. Status code: {}.\n {} \n errors: {}" .format(
-        response.status_code,
-        gql_query_str,
-        response.json()['errors']))
-
-def __result_has_gql_errors__(response):
-  return 'errors' in response.json().keys()
+    if __exist_not_empty_result(response):
```
IvanIvanoff: I am not convinced that everything except non-empty result is an error. For example if you want to fetch the top transactions in an interval and there are none - the correct result is empty.

Also, inverting the checks is not correct, too. It can happen that you have `data` *and* `errors` in the same response. In such case I think we need to have the errors check first and do not ignore the errors

san/graphql.py
```diff
@@ -5,45 +5,47 @@
...
-        gql_query_str,
-        response.json()['errors']))
-  elif __exist_not_empty_result(response):
-    return response.json()['data']
-  else:
-    raise SanError(
-      "Error running query. Status code: {}.\n {} \n errors: {}" .format(
-        response.status_code,
-        gql_query_str,
-        response.json()['errors']))
```
IvanIvanoff: Removing only this line here should fix the error handling (there is no `errors` here) - no further restructuring is needed.

tspenov:
@spiderjako what python formatter do you use ? You should commit the configuration also so we start using it

san/sanbase_graphql_helper.py
```diff
@@ -274,9 +274,7 @@ def _format_to_date(datetime_obj_or_str):
     if isinstance(datetime_obj_or_str, datetime.datetime):
         datetime_obj_or_str = datetime_obj_or_str.isoformat()
 
-    dt = iso8601.parse_date(datetime_obj_or_str) + \
-        datetime.timedelta(hours=23, minutes=59, seconds=59)
-    return dt.isoformat()
+    return iso8601.parse_date(datetime_obj_or_str)
```
tspenov: that is from other PR :?

spiderjako: Yes my mistake.

san/graphql.py
```diff
@@ -5,45 +5,47 @@
...
-        response.status_code,
-        gql_query_str))
+    headers = {}
+    if ApiConfig.api_key:
+        headers = {'authorization': "Apikey {}".format(ApiConfig.api_key)}
+
+    response = requests.post(
+        SANBASE_GQL_HOST,
+        json={
+            'query': gql_query_str},
```
IvanIvanoff: what formatter tool resulted in this thing?

spiderjako: autopep8, guess it didn't go through this one

PR #95, opened 2019-12-18
Fix to_date parameter

spiderjako:
The following script:
```python
get(
    "transaction_volume/santiment",
    from_date="2018-01-01T00:00:00Z",
    to_date="2018-01-01T01:00:00Z",
    interval="5m"
)
```
Returns 300 records, from 1st to 2nd of January, 2018, which is false.
It should return only records for the first hour of 01.01.2018.
Problem was an implementation detail, which adds an additional 23:59:59 hours to the to_date.

### DIFF
--- a/san/sanbase_graphql_helper.py
+++ b/san/sanbase_graphql_helper.py
@@ -1,5 +1,6 @@
 import iso8601
 import datetime
+import re
 
 _DEFAULT_INTERVAL = '1d'
 _DEFAULT_SOCIAL_VOLUME_TYPE = 'PROFESSIONAL_TRADERS_CHAT_OVERVIEW'
@@ -273,9 +274,13 @@ def _format_from_date(datetime_obj_or_str):
 def _format_to_date(datetime_obj_or_str):
     if isinstance(datetime_obj_or_str, datetime.datetime):
         datetime_obj_or_str = datetime_obj_or_str.isoformat()
+    else:
+        if _validate_iso8601(datetime_obj_or_str):
+            dt = iso8601.parse_date(datetime_obj_or_str)
+        else:
+            dt = iso8601.parse_date(datetime_obj_or_str) + \
+                datetime.timedelta(hours=23, minutes=59, seconds=59)
 
-    dt = iso8601.parse_date(datetime_obj_or_str) + \
-        datetime.timedelta(hours=23, minutes=59, seconds=59)
     return dt.isoformat()
 
 def _format_all_return_fields(fields):
@@ -287,3 +292,13 @@ def _format_return_fields(fields):
     return list(map(
         lambda el: el[0] + '{{' + ' '.join(el[1]) + '}}' if isinstance(el, tuple) else el
     , fields))
+
+def _validate_iso8601(str_val):
+    regex = r'^(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9]):([0-5][0-9]):([0-5][0-9])(\.[0-9]+)?(Z|[+-](?:2[0-3]|[01][0-9]):[0-5][0-9])?$'
+    match_iso8601 = re.compile(regex).match
+    try:            
+        if match_iso8601( str_val ) is not None:
+            return True
+    except:
+        pass
+    return False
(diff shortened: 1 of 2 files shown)


san/sanbase_graphql_helper.py
```diff
@@ -274,9 +274,7 @@ def _format_to_date(datetime_obj_or_str):
     if isinstance(datetime_obj_or_str, datetime.datetime):
         datetime_obj_or_str = datetime_obj_or_str.isoformat()
 
-    dt = iso8601.parse_date(datetime_obj_or_str) + \
-        datetime.timedelta(hours=23, minutes=59, seconds=59)
-    return dt.isoformat()
+    return iso8601.parse_date(datetime_obj_or_str).isoformat()
```
IvanIvanoff: This will break a lot of old behaviour

spiderjako: You're right, going to change it up

san/sanbase_graphql_helper.py
```diff
@@ -273,9 +273,13 @@ def _format_from_date(datetime_obj_or_str):
 def _format_to_date(datetime_obj_or_str):
     if isinstance(datetime_obj_or_str, datetime.datetime):
         datetime_obj_or_str = datetime_obj_or_str.isoformat()
+    else:
+        if 'T' in datetime_obj_or_str and 'Z' in datetime_obj_or_str:
+            dt = iso8601.parse_date(datetime_obj_or_str)
```
IvanIvanoff: Google how to validate iso8601 datetime formats

IvanIvanoff: Instead of parsing datetime and treating it as date in the else block, do the opposite - try parsing date and treat as datetime in the else block
```python
date_format = '%Y-%m-%d'
try:
  date = datetime.datetime.strptime(date_str, date_format)
  # treat as date, old behaviour
except ValueError, e:
  # treat as datetime, new behavior
```

san/sanbase_graphql_helper.py
```diff
@@ -271,11 +271,17 @@ def _format_from_date(datetime_obj_or_str):
 
 
 def _format_to_date(datetime_obj_or_str):
-    if isinstance(datetime_obj_or_str, datetime.datetime):
-        datetime_obj_or_str = datetime_obj_or_str.isoformat()
+    date_format = '%Y-%m-%d'
+    if not isinstance(datetime_obj_or_str, datetime.datetime):
```
IvanIvanoff: instead of `if not` just use `if` and swap the bodies of the if and else

san/sanbase_graphql_helper.py
```diff
@@ -271,11 +271,17 @@ def _format_from_date(datetime_obj_or_str):
 
 
 def _format_to_date(datetime_obj_or_str):
-    if isinstance(datetime_obj_or_str, datetime.datetime):
-        datetime_obj_or_str = datetime_obj_or_str.isoformat()
+    date_format = '%Y-%m-%d'
+    if not isinstance(datetime_obj_or_str, datetime.datetime):            
+        try:
+            datetime.datetime.strptime(datetime_obj_or_str, date_format)
```
IvanIvanoff: date_format is used only here - you can just use
```python
datetime.datetime.strptime(datetime_obj_or_str, '%Y-%m-%d')
```

san/sanbase_graphql_helper.py
IvanIvanoff: Actually, it can be better handled with an early return:
```python
def _format_to_date(datetime_obj_or_str):
    return datetime_obj_or_str.isoformat()

  try:
    # Throw if the string is not date-formated, parse as date otherwise
    datetime.datetime.strptime(datetime_obj_or_str, date_format)
    dt = iso8601.parse_date(datetime_obj_or_str) + datetime.timedelta(hours=23, minutes=59, seconds=59)
  except:
      dt = iso8601.parse_date(datetime_obj_or_str)

  return dt.isoformat()
```

otherwise you're nesting `try/except` in an `if not` case which is hard to read

spiderjako: Thanks, that's better really

san/sanbase_graphql_helper.py
spiderjako: Yeah, you're right

PR #98, opened 2020-01-10
Removing news documentation from README.md

spiderjako:


san/sanbase_graphql.py
```diff
@@ -223,6 +223,7 @@ def eth_top_transactions(idx, slug, **kwargs):
 
 
 def news(idx, tag, **kwargs):
+    print('WARNING! This metric is going to be removed soon')
```
IvanIvanoff: removed in version X.Y.Z - `soon` does not mean anything

PR #101, opened 2020-01-29
Event study

JanSmirny:
Adding event study to sanpy


san/eventstudy.py
```diff
@@ -0,0 +1,277 @@
...
+    bench_prices = np.reshape(bench_prices,len(bench_prices))
+    stock_prices = np.reshape(stock_prices,len(stock_prices))
+    if len(stock_prices) == 0:
+        return None
+    m, b = np.polyfit(bench_prices, stock_prices, 1)
+    return m
+
+# Simple event study
+
+def eventstudy(data, ev_data,starting_point = 30, benchmark='bitcoin',
```
IvanIvanoff: that's one big function over 200+ lines - needs refactoring. Also let's use snake case for naming functions to be consistent

san/eventstudy.py
```diff
@@ -0,0 +1,277 @@
...
+    aligned_prices = bench_prices.align(stock_prices,join='inner')
+    bench_prices = aligned_prices[0]
+    stock_prices = aligned_prices[1]
+    bench_prices = np.array( bench_prices.values )
+    stock_prices = np.array( stock_prices.values )
+    bench_prices = np.reshape(bench_prices,len(bench_prices))
+    stock_prices = np.reshape(stock_prices,len(stock_prices))
+    if len(stock_prices) == 0:
+        return None
+    m, b = np.polyfit(bench_prices, stock_prices, 1)
```
IvanIvanoff: please use better var names here

san/eventstudy.py
```diff
@@ -0,0 +1,277 @@
+import numpy as np
+import pandas as pd
+import seaborn as sns
```
IvanIvanoff: This dependency is missing in the requirements.txt

kalkov: It is not used at all. Removed.

examples/extras/event_study.py
```diff
@@ -0,0 +1,32 @@
+from datetime import date
+import numpy as np
+import san
+
+import sys
+sys.path.append("../..")
```
IvanIvanoff: 😨

examples/extras/event_study.py
```diff
@@ -0,0 +1,32 @@
...
+data["buy"] = (data["daa_performance"] < (data["sd_rolling"] * (-1) * 2))
+# Sell Signal: Is the current trend higher than two standard deviations?
+data["sell"] = (data["daa_performance"] > (data["sd_rolling"] * 2))
+
+
+# Helper function to get the signals in the right format:
+signals = signals_format(data["sell"],'ETH_close')
+
+# Calling the event study:
+event_study(price, signals, starting_point = 30)
```
IvanIvanoff: The code here is not organized in functions/classes?

san/extras/event_study.py
```diff
@@ -0,0 +1,286 @@
...
+
+            returns.append(_return)
+            benchmark_returns.append(benchmark_return)
+            sample_size += 1
+
+            """
+            Calculate beta by getting the last X days of data
+            1. Create a DataFrame containing the data for the necessary sids within that time frame
+            2. Pass that DataFrame into our calc_beta function in order to spit out a beta
+            """
```
IvanIvanoff: Should be separate function

san/extras/event_study.py
```diff
@@ -0,0 +1,286 @@
...
+
+    day_numbers = build_day_numbers(starting_point)
+
+    for day_num in day_numbers:
+        returns = []
+        benchmark_returns = []
+        abnormal_returns = []
+        sample_size = 0
+
+        for date, row in ev_data.iterrows():
```
IvanIvanoff: The whole body of this for (or its parent for?) should probably be a separate function. Here we have nested `if`s and `for`s which is not readable/maintainable

examples/extras/event_study.py
kalkov: Open for suggestions how to write an example for something that is not released yet.

examples/extras/event_study.py
JanSmirny: This is not part of the module but the example on how to prepare signals and evaluate them via event study. Its meant to be run in a jupyter notebook. So I dont know if this needs to be modular. But I can do it of you say its more usable that way.

examples/extras/event_study.py
valo: It should be possible to install the module using `pip` from a local folder, so it should be possible to make the notebook work as if the module is published officially.

examples/extras/event_study.py
valo: May be this should be converted to a Jupiter notebook?

san/extras/event_study.py
valo: Yeah. This function is way too long. Let's try to break it down into logical pieces.

examples/extras/event_study.py
JanSmirny: Jup makes sense

san/extras/event_study.py
JanSmirny: Alright. Ricardo changed it to be separate functions. Have a look.

PR #103, opened 2020-02-05
Add extras module

kalkov:
The aim  is to separate scripts that need extra dependencies and are not related to graphQL API.
Have introduced [pipenv](https://github.com/pypa/pipenv) for easier development environment management.


valo:
Can we also attach a Jupiter notebook with the MA example? This will allow to see the graphs we generate in the summary directly in GitHub

README.md
```diff
@@ -1265,7 +1276,37 @@ Example result:
...
 ```
 {'defaultAggregation': 'AVG', 'metric': 'nvt'}
+
 ```
+
+## Extras
+
+### Backtest
+
+TBD
```
spiderjako: Shouldn't we add some info here?

Pipfile
```diff
@@ -0,0 +1,10 @@
+[[source]]
+url = "https://pypi.python.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+sanpy = {editable = true, path = "."}
+
+[requires]
+python_version = "3.6"
```
tspenov: What does it mean `requires` python 3.6  in this case ? If I have only 3.7 or 3.8
will it be ok. Generally we need something like >= 3.6 (or >= 3.0 if we want to support all python 3)

README.md
```diff
@@ -1265,7 +1276,37 @@ Example result:
...
+
+## Development
+
+It is recommended to use [pipenv](https://github.com/pypa/pipenv) for managing your local environment.
+
+Setup project:
+
+```
+pipenv install
```
tspenov: I generally prefer `asdf` as version manager because it can be used also for elixir and other language envs we have. But I guess we want be able to have separate dependancies that way :?

Pipfile
kalkov: The file will be used only for development, `pipenv` will create a virtualenv and install the required python for you. The idea is to have the minimal supported version locked so everyone is using the same one and not accidentally introduce code that will not work on that version.

kalkov: Here is a nice article about pipenv: http://totempaal.tj/2018/07/09/python-dependency-management.html

README.md
kalkov: `pipenv` is considered the pythonic way, it is developed under [Python Packaging Authority](https://github.com/pypa).

Pipfile
tspenov: @kalkov thanks for the link

README.md
tspenov: lets add info or remove from README

kalkov: Removed the backtest section and left the extras section with a link to examples folder.

kalkov:
@valo Added example jupyter notebook.

PR #107, opened 2020-03-19
Add metrics to v2

spiderjako:


setup.py
```diff
@@ -6,7 +6,7 @@
 
 setuptools.setup(
     name="sanpy",
-    version="0.7.15",
+    version="0.7.17",
```
IvanIvanoff: why 2 versions?

spiderjako: I was looking at the PR and the version was 0.7.16, it seems that it is actually 0.7.17 at the moment. I bumped it to 0.7.18 now

on-chain-metrics.md
```diff
@@ -15,6 +15,11 @@ List of with metrics with 1 day minimal resolution:
 - daily_opening_price_usd
 - daily_trading_volume_usd
 - daily_active_addresses
+- active_deposits
+- active_withdrawals
+- withdrawal_transactions
+- exchange_token_supply
```
IvanIvanoff: No such metric is present already

san/tests/test_sanbase_integration.py
```diff
@@ -541,7 +534,15 @@ def test_available_metrics():
...
-        'nvt'
+        'nvt',
+        'nvt_transaction_volume',
+        'network_growth',
+        'mean_dollar_invested_age',
+        'active_deposits',
+        'active_withdrawals',
+        'withdrawal_transactions',
+        'exchange_token_supply',
+        'percent_of_total_supply_on_exchanges'
```
IvanIvanoff: amount_in_top_holders is also present now

PR #108, opened 2020-05-14
Event Study Warning Fixed

RicardoSalaz:
In this version, the warnings of the previous version were corrected and a data cleaning function was added.


tzanko-matev:
Looks good to me

PR #110, opened 2020-06-09
Adding missing metrics to getMetric

spiderjako:
### Summary
* Changing the way SanPy checks for whether a metric is in getMetric or not
* Addition of available_metrics_for_slug - function to get all the metrics for a given slug
* Fix for the ohlcv function (currently raises and error because it tries to fetch columns from an empty df)


san/v2_metrics_list.py
```diff
@@ -73,5 +73,87 @@
       "withdrawal_transactions",
       "supply_on_exchanges",
       "percent_of_total_supply_on_exchanges",
-      "amount_in_top_holders"
+      "amount_in_top_holders",
```
tspenov, line 76: can you remind me why we list these manually instead of fetching from the api like this ?
```
{
  getAvailableMetrics
}
```

spiderjako: If I remember correctly, it had something to do with the tests and the fact that you can't mock 2 responses in one test, otherwise it worked

on-chain-metrics.md
```diff
@@ -124,4 +124,114 @@ List of metrics with 5 minutes minimal resolution:
 - age_destroyed
 - exchange_inflow
 - exchange_outflow
-- exchange_balance
\ No newline at end of file
+- exchange_balance
```
tspenov, line 127: I think we should also remove this file at some point and point to the query for available metrics + (list only those that are not in available metrics). But lets leave for now.

san/sanbase_graphql_helper.py
```diff
@@ -6,11 +6,7 @@
 _DEFAULT_SOURCE = 'TELEGRAM'
 _DEFAULT_SEARCH_TEXT = ''
 
-_QUERY_MAPPING = {
-    'daily_active_addresses': {
```
tspenov, line 10: why remove `daily_active_addresses` ?

san/slug_metrics.py
```diff
@@ -0,0 +1,13 @@
+from .graphql import execute_gql
+
+def slug_metrics(slug):
```
tspenov, line 3: available_metrics_for_slug ?

san/graphql.py
```diff
@@ -38,6 +38,14 @@ def __handle_success_response__(response, gql_query_str):
                 gql_query_str))
 
 
+def get_all_available_metrics():
```
tspenov, line 41: Can you reuse `execute_gql` here ? Also I think this function does not belong to this file but to file with other queries.

san/slug_metrics.py
IvanIvanoff: came here to suggest the same

IvanIvanoff:
@spiderjako Please bump the sanpy version in the PR

san/available_metrics_for_slug.py
```diff
@@ -0,0 +1,13 @@
+from .graphql import execute_gql
```
tspenov, line 1: I think `available_metrics_for_slug ` that belongs to `available_metrics.py`.
Also you should add it into the README below `Available Metrics` so users know about it.

san/sanbase_graphql_helper.py
spiderjako: Because it currently uses the daily_active_addresses from getMetric and not the dailyActiveAddresses API

tspenov: It returns ['datetime', 'activeAddresses']  while the getMetric on returns ['datetime', 'value']. That will break the code of those currently using it. 
Isn't it the same case with `network_growth`, `dev_activity` , etc but they are not removed ?
Also didn't we have a message that we will deprecated these that are not using `getMetric` in V. 0.9.0 ?

spiderjako: The PR just got a little streched out and I forgot about it, the main reason why I removed it is because there is no daily_active_addresses function in the first place, maybe we've agreed to deprecate the old API for the getMetric some time back, don't really know why it isn't here now. Another reason is that the 'get' function currently favours the getMetric version more over the old API one, the "switch" that we've made in this PR will change that and will break stuff anyways, but you're right about the case with network_growth and dev_activity - we should decide on whether we remove these same-named old APIs altogether or leave all of them be, I think we should remove them all personally

tspenov: Ok, got it. In that case we should find all queries like `network_growth`, etc that are in both getMetric and without and remove them also. Because currnetly if you call `network_growth`, you get `value` field. After this PR you'll get `newAddresses` - so that breaks existing code. If we remove also `network growth` - it will go to getMetric and you'll get `value` again.
So we should remove all that are also in getMetric.

tspenov: And also update the README that it returns `value` cause I guess we've missed that previous time.

spiderjako: Done, also added some more metrics to the to-be-deprecated metrics list

on-chain-metrics.md
```diff
@@ -124,4 +124,114 @@ List of metrics with 5 minutes minimal resolution:
...
+- sentiment_negative_total,
+- sentiment_balance_telegram,
+- sentiment_balance_reddit,
+- sentiment_balance_bitcointalk,
+- sentiment_balance_twitter,
+- sentiment_balance_discord,
+- sentiment_balance_professional_traders_chat,
+- sentiment_balance_total
+
+### Bitmex
```
IvanIvanoff, line 160: `Derivatives` not `Bitmex`

on-chain-metrics.md
```diff
@@ -124,4 +124,114 @@ List of metrics with 5 minutes minimal resolution:
...
+
+- stock_to_flow
+
+### Supply Outside Exchanges
+
+- supply_outside_exchanges
+
+### Amount in Top Holders
+
+- amount_in_top_holders
```
IvanIvanoff, line 194: Can you revisit the list and groupings?
- Why `supply_on_exchanges` is removed? There is no reason to have a separate `Supply Outside Exchanges` group just to put one metric there. 
- There are more top holders metrics. Again - no point in having a group with the same name as the single metric inside it. We have amount in exchange, non_exchange holders, holder distributions, etc. etc.

on-chain-metrics.md
```diff
@@ -124,4 +124,114 @@ List of metrics with 5 minutes minimal resolution:
...
+
+### Maker Dao
```
IvanIvanoff, line 196: `MakerDAO`

on-chain-metrics.md
```diff
@@ -124,4 +124,114 @@ List of metrics with 5 minutes minimal resolution:
...
+- price_usd_change_30d,
+- price_usd_change_7d,
+- price_usd_change_1d,
+- volume_usd_change_30d,
+- volume_usd_change_7d,
+- volume_usd_change_1d
+
+### Dev Activity
+
+- dev_activity
```
IvanIvanoff, line 182: github_activity as well?

IvanIvanoff: Also:
- dev_activity_contributors_count
- github_activity_contributors_count

on-chain-metrics.md
```diff
@@ -124,4 +121,140 @@ List of metrics with 5 minutes minimal resolution:
...
+- volume_usd_change_30d
+- volume_usd_change_7d
+- volume_usd_change_1d
+
+### Github/Dev Activity
+
+- github_activity
+- github_activity_contributors_count
+- dev_activity
+- dev_activity_1d
```
IvanIvanoff, line 202: let's not include here as it will only confuse

on-chain-metrics.md
```diff
@@ -124,4 +121,140 @@ List of metrics with 5 minutes minimal resolution:
...
+- dev_activity_contributors_count
+- dev_activity_change_30d
+
+### Stock to Flow
+
+- stock_to_flow
+
+### Supply Exchanges
```
IvanIvanoff, line 210: Sounds strange.

PR #112, opened 2020-06-19
Add timeseries data complexity

spiderjako:


README.md
```diff
@@ -1256,6 +1257,25 @@ Example result:
 
 ```
 
+### Metric Complexity
+
+Fetch the complexity of a query
```
IvanIvanoff, line 1262:
```suggestion
Fetch the complexity of a metric. The complexity depends on the from/to/interval parameters, as well as the metric and the subscription plan. A request might have a maximum complexity of 20000. If a request has a higher complexity there are a few ways to solve the issue:
- Break down the request into multiple requests with smaller from-to ranges.
- Upgrade to a higher subscription plan.



```

README.md
```diff
@@ -795,6 +796,25 @@ datetime                              slug     change    status
...
+
+```python
+san.metric_complexity(
+    metric='price_usd',
+    from_datetime='2020-02-02',
```
tspenov, line 806:
```suggestion
    from_date='2020-02-01',
```

README.md
```diff
@@ -795,6 +796,25 @@ datetime                              slug     change    status
...
+    to_datetime='2020-03-20',
```
tspenov, line 807:
```suggestion
    to_date='2020-02-20',
```

README.md
```diff
@@ -795,6 +796,25 @@ datetime                              slug     change    status
...
+    from_date='2020-02-01',
+    to_date='2020-02-20',
+    interval='1d'
+)
+```
+
+Example result:
+
+```
+29
```
tspenov, line 815: Also the result should be different.

setup.py
```diff
@@ -6,7 +6,7 @@
 
 setuptools.setup(
     name="sanpy",
-    version="0.7.19",
+    version="0.7.20",
```
tspenov, line 9: I suggest bumping to `0.8.0`

PR #119, opened 2020-07-13
Adding available since to sanpy

spiderjako:


README.md
```diff
@@ -818,6 +819,19 @@ Example result:
 31
 ```
 
+### Available Since
+
+Fetch the first datetime in which a metric is available for a given slug.
```
IvanIvanoff, line 824:
```suggestion
Fetch the first datetime for which a metric is available for a given slug.
```

README.md
```diff
@@ -30,6 +30,7 @@ Santiment API python client.
     - [Emerging Trends](#emerging-trends)
     - [Top Social Gainers Losers](#top-social-gainers-losers)
     - [Metric Complexity](#metric-complexity)
+    - [Available Since](#available-since)
```
IvanIvanoff, line 33: This should not be nested under `metrics` (as well as the metric complexity) I guess?

PR #123, opened 2020-11-04
Add support for utc_now in sanpy

spiderjako:


IvanIvanoff (approved):
Bump the version

san/sanbase_graphql_helper.py
```diff
@@ -249,13 +249,19 @@ def _default_from_date():
 
 
 def _format_from_date(datetime_obj_or_str):
+    if isinstance(datetime_obj_or_str, str) and 'utc_now' in datetime_obj_or_str:
```
tspenov, line 252: I think the check should be more precise - here we check only if `utc_now` string is present but it can still be invalid. Better use a regex in this case.

tspenov:
I think a note to documentation should be added that now `utc_now` and `utc_now - 1d` expressions can be used now for from, to dates.

san/sanbase_graphql_helper.py
spiderjako: @IvanIvanoff What do you think about regex-es? I know you that kind of oppose them :D

IvanIvanoff: We don't need to replicate all the checks - the API will do the work. Here we can do only some basic checks like this one. The API error will be properly reported here and we would not need to maintain and add future changes that are introduced in the API types.

tspenov: Ok, lets leave the proper check to the API

PR #136, opened 2021-11-23
Adding api call remainder function

spiderjako:
<Description>
Adding a function for returning a list of API calls made by date:

```
[
    ('2021-11-19T00:00:00Z', 1),
    ('2021-11-22T00:00:00Z', 2),
    ('2021-11-23T00:00:00Z', 3)
]
```
...
and a function to return a dict of remaining API calls:
```
{
  'month_remaining': '1000',
  'hour_remaining': '500',
  'minute_remaining': '100'
}
```


IvanIvanoff:
Can we also have a function that returns how many calls are left until rate limits are applied?

PR #143, opened 2022-05-29
Do not pin pandas version in setup.py

IvanIvanoff:


IvanIvanoff:
Closed by https://github.com/santiment/sanpy/pull/144

PR #144, opened 2022-05-30
Non-hardcoded versions

spiderjako:


setup.py
```diff
@@ -22,7 +22,7 @@
         'Cython'
     ],
     install_requires=[
-        'pandas==1.3.0',
+        'pandas>=1.3.0',
```
IvanIvanoff, line 25: How does `>=` work? Does it mean that only the last digit in the version can change (i.e. from 1.3.0 to 1.3.99, but not to 1.4.0) or all versions above are accepted?

spiderjako: I just wanted to test it out this way, had no idea that it'd work lol

setup.py
```diff
@@ -18,11 +18,11 @@
     tests_require=['nose'],
     packages=setuptools.find_packages(),
     setup_requires=[
-        'numpy==1.21.1',
+        'numpy==1.22.0',
```
IvanIvanoff, line 21: This is still hardcoded.

Pipfile
```diff
@@ -7,4 +7,4 @@ name = "pypi"
 sanpy = {editable = true, path = "."}
 
 [requires]
-python_version = "3.6"
+python_version = "3.9"
```
IvanIvanoff, line 10: isn't 3.8 enough?

Jenkinsfile
```diff
@@ -1,5 +1,5 @@
 podTemplate(label: 'sanpy-builder', containers: [
-    containerTemplate(name: 'python', image: 'python:3-stretch', command: 'cat', ttyEnabled: true)
+    containerTemplate(name: 'python', image: 'python:3.9-buster', command: 'cat', ttyEnabled: true)
```
IvanIvanoff, line 2: maybe 3.8 so we don't accidentally introduce a 3.9 requirement

IvanIvanoff (approved):

PR #151, opened 2022-09-28
Add async batch

IvanIvanoff:
Multiple queries can be executed in a batch to speed up the performance.

There are two batch classes provided - `Batch` and `AsyncBatch`.

- `Batch` combines all the provided queries in a single GraphQL document and
executes them in a single HTTP request. This batching technique should be used
when lightweight queries that don't fetch a lot of data are used. The reason is
that the [complexity](https://academy.santiment.net/for-developers/#graphql-api-complexity) of each query
is accumulated and the batch can be rejected.
- `AsyncBatch` executes all the queries in separate API calls. The benefit of using `AsyncBatch`
  over looping and executing every API call is that the queries can be executed concurrently.
  The concurrency is controlled by the `max_workers` optional parameter to the `execute` function.
  By default the `max_workers` value is 10.
  
Note: If you have been using `Batch()` and want to switch to the newer `AsyncBatch()` you only need to
change the batch initialization. The functions for adding queries and executing the batch, as well as the
format of the response, are the same.

```python
from san import AsyncBatch

batch = AsyncBatch()

batch.get(
    "daily_active_addresses/santiment",
    from_date="2018-06-01",
    to_date="2018-06-05",
    interval="1d"
)
batch.get(
    "transaction_volume/santiment",
    from_date="2018-06-01",
    to_date="2018-06-05",
    interval="1d"
)
[daa, trx_volume] = batch.execute(max_workers=10)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,14 +242,12 @@ batch.get(
     to_date="2018-06-05",
     interval="1d"
 )
-
 batch.get(
     "transaction_volume/santiment",
     from_date="2018-06-01",
     to_date="2018-06-05",
     interval="1d"
 )
-
 [daa, trx_volume] = batch.execute(max_workers=10)
 ```
 
--- a/README.md
+++ b/README.md
@@ -191,6 +191,24 @@ Example result:
 
 ### Batching multiple queries
 
+Multiple queries can be executed in a batch to speed up the performance.
+
+There are two batch classes provided - `Batch` and `AsyncBatch`.
+
+- `Batch` combines all the provided queries in a single GraphQL document and
+executes them in a single HTTP request. This batching technique should be used
+when lightweight queries that don't fetch a lot of data are used. The reason is
+that the [complexity](https://academy.santiment.net/for-developers/#graphql-api-complexity) of each query
+is accumulated and the batch can be rejected.
+- `AsyncBatch` executes all the queries in separate API calls. The benefit of using `AsyncBatch`
+  over looping and executing every API call is that the queries can be executed concurrently.
+  The concurrency is controlled by the `max_workers` optional parameter to the `execute` function.
+  By default the `max_workers` value is 10.
+  
+Note: If you have been using `Batch()` and want to switch to the newer `AsyncBatch()` you only need to
+change the batch initialization. The functions for adding queries and executing the batch, as well as the
+format of the response, are the same.
+
 ```python
 from san import Batch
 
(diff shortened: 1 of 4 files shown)


robcaulk:
This is a very needed feature - thank you for implementing.

san/async_batch.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self):
+        self.queries = []
+
+    def get(self, dataset, **kwargs):
+        self.queries.append([dataset, kwargs])
+
+    def execute(self, max_workers=10):
+        graphql_result = {}
+
+        with ProcessPoolExecutor(max_workers=max_workers) as executor:
```
apozdeyev, line 37: for network requests it is better to use ThreadPoolExecutor, should be less overhead

README.md
```diff
@@ -213,6 +231,28 @@ batch.get(
...
+)
+
+batch.get(
+    "transaction_volume/santiment",
+    from_date="2018-06-01",
+    to_date="2018-06-05",
+    interval="1d"
+)
+
+[daa, trx_volume] = batch.execute(max_workers=10)
```
spiderjako, line 253:
```suggestion
from san import AsyncBatch

def main():
    batch = AsyncBatch()

    batch.get(
        "daily_active_addresses/santiment",
        from_date="2018-06-01",
        to_date="2018-06-05",
        interval="1d"
    )
    batch.get(
        "transaction_volume/santiment",
        from_date="2018-06-01",
        to_date="2018-06-05",
        interval="1d"
    )
    [daa, trx_volume] = batch.execute(max_workers=10)

    return [daa, trx_volume]

if __name__ == '__main__':
    main()
```

IvanIvanoff: Already updated, good finding. I've also added some explanation about that.

san/async_batch.py
IvanIvanoff: I've changed that.

README.md
IvanIvanoff: Actually, after changing from `ProcessPoolExecutor` to `ThreadPoolExecutor` it seems that this is no longer needed.

san/async_batch.py
```diff
@@ -0,0 +1,56 @@
...
+                
+                graphql_result[key] = value
+            
+            result = self.__transform_batch_result(graphql_result)
+            return result
+
+
+    def __transform_batch_result(self, graphql_result):
+        result = []
+        idxs = sorted([int(k.split('_')[1]) for k in graphql_result.keys()])
```
apozdeyev, line 50: you can omit square brackets here

apozdeyev (approved):

PR #163, opened 2022-11-04
Allow configuring sanpy apikey through the env

IvanIvanoff:
## Configuration

You can provide an API key which gives access to the restricted metrics in two different ways:

### Read the API key from the environment

During loading of the `san` module, if the `SANPY_APIKEY` exists, its content
is read and set as the API key.

```shell
export SANPY_APIKEY="my_apikey"
```
```python
import san
>>> san.ApiConfig.api_key
'my_apikey'
```

### Manually configure an API key

```python
import san
san.ApiConfig.api_key = "my_apikey"
```

### How to obtain an API key

To obtain an API key you should [log in to sanbase](https://app.santiment.net/login)
and go to the `Account` page - [https://app.santiment.net/account](https://app.santiment.net/account).
There is an `API Keys` section and a `Generate new api key` button.


tspenov:
There is also this file for environment variables: https://github.com/santiment/sanpy/blob/master/san/env_vars.py

apozdeyev (approved):

PR #176, opened 2023-12-15
Create __init__.py

mrchaos:
add __init__.py
For "from san.extras.strategy.strategy import Strategy"


IvanIvanoff:
@mrchaos thanks for the contribution!
We've merged the PR and the change is available in the new 0.11.8 version: https://pypi.org/project/sanpy/0.11.8/

PR #181, opened 2024-03-20
Remove old and unused functions

IvanIvanoff:
These functions have a getMetric equvialent or are deprecated

### DIFF
--- a/san/sanbase_graphql.py
+++ b/san/sanbase_graphql.py
@@ -4,94 +4,20 @@ from san.batch import Batch
 from san.error import SanError
 
 
-# to be removed
-
-
-def burn_rate(idx, slug, **kwargs):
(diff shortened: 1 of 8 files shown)


san/sanbase_graphql.py
```diff
@@ -307,27 +166,6 @@ def emerging_trends(idx, **kwargs):
     return query_str
 
 
-def top_social_gainers_losers(idx, **kwargs):
```
tspenov, line 310: You should check also to remove in the README. For example this one is present there.

IvanIvanoff: Done.

spiderjako (approved):

tspenov (approved):

PR #187, opened 2025-04-16
Use the JSON getMetric endpoints for increased performance

IvanIvanoff:


IvanIvanoff:
@tspenov Ruff is used for linting so far, I will open a new PR and run `ruff format` so we can have consistent formatting in the whole project.

PR #199, opened 2026-04-06
WIP: introduce client-based core API

createMonster:
## Summary
- introduce `SanClient` and `ClientConfig` as the new client-based core
- route module-level APIs through a default client to preserve backward compatibility
- add coverage for explicit client usage and the temporary global config compatibility path

## Validation
- `pytest san/tests -q`


createMonster:
Closing this draft because it was based on an outdated local master and bypassed the transport-based architecture introduced in #195. I will reopen Phase 2 from the latest master in a fresh WIP PR.

